PR #14, opened 2020-12-16
Seas new

uphoffc:


app/common/PetscLinearSolver.h
```diff
@@ -0,0 +1,59 @@
...
+                                                topo.numGhostNeighbours(), topo.comm());
+        b_ = std::make_unique<PetscBlockVector>(dgop.block_size(), topo.numLocalElements(),
+                                                topo.comm());
+        x_ = std::make_unique<PetscBlockVector>(*b_);
+        dgop.assemble(*A_);
+        dgop.rhs(*b_);
+
+        CHKERRTHROW(KSPCreate(topo.comm(), &ksp_));
+        CHKERRTHROW(KSPSetType(ksp_, KSPCG));
+        CHKERRTHROW(KSPSetOperators(ksp_, A_->mat(), A_->mat()));
```
hpc4geo, line 33: @uphoffc Hard wiring that these two matrices is very limiting. Can you at least provide a new getter (e.g. A_->pcmat()) and have the default implementation of pcmat() return the same operator as mat()?

uphoffc: A_ is a PetscBlockMatrix which wraps a "Mat" into a C++ object. A_ itself is defined a few lines above and stems from a DGOperator class, which itself provides a standard interface for getting some info about the block structure and provides an "assemble" function.
Therefore, I believe one could optionally pass another DGOperator to the constructor which would assemble a matrix used for constructing the preconditioner.
The question here is whether the matrix used for the preconditioner is going to have the same block structure as the operator matrix or whether it makes sense that it is something completely different?

hpc4geo: It should not be assumed that the matrix used to define the preconditioner will have the same MatType (which I think here is BAIJ) as that of the operator. For example, many preconditioners only support MatType AIJ (SEQAIJ or MPIAIJ). Another use case would be if you defined a matrix-free operator for A (possibly represented as MatShell), but you wish to assemble something for the preconditioner (e.g. MATAIJ).

So it would seem to be that operator associated with the "Bmat" arg in KSPSetOperators should be assumed to be something completely different.

uphoffc: Ah I see. Actually PetscBlockMatrix supports AIJ as well as BAIJ, so you should be able to switch the type with -mat_type. (is there actually an option like -pc_mat_type or so?)

Moreover, "assemble" works for AIJ as well as for BAIJ matrices, therefore I just need to call assemble twice when the mat type is different.

app/common/PetscLinearSolver.h
```diff
@@ -0,0 +1,59 @@
...
+        A_ = std::make_unique<PetscBlockMatrix>(dgop.block_size(), topo.numLocalElements(),
+                                                topo.numLocalNeighbours(),
+                                                topo.numGhostNeighbours(), topo.comm());
+        b_ = std::make_unique<PetscBlockVector>(dgop.block_size(), topo.numLocalElements(),
+                                                topo.comm());
+        x_ = std::make_unique<PetscBlockVector>(*b_);
+        dgop.assemble(*A_);
+        dgop.rhs(*b_);
+
+        CHKERRTHROW(KSPCreate(topo.comm(), &ksp_));
```
hpc4geo, line 31: @uphoffc It would be very useful to allow one to attach prefix to the KSP via KSPSetOptionsPrefix() here. Thus enabling command line args to define the configuration of a specific KSP object. -poisson_ksp_type cg -elasticity_ksp_type gmres

app/common/PetscLinearSolver.h
hpc4geo: If it supports both why the hell is the class called PetscBlockMatrix? Wouldn't PetscMatrix be less ambiguous? 

Incidentally, you can create a matrix with type MATAIJ, set its block size (via MatSetBlockSize) and assemble into via MatSetValues, MatSetValuesBlocked. You would also be better (performance wise) calling MatSetValuesBlockedLocal(). I fail to see the true advantage of this class (or PetscBlockVector). 

The name PetscBlockVector also seems odd. Maybe a name reflecting that the matrix is associated with a DG function defined over a mesh space might be more appropriate. Maybe BilinearForm? Maybe TandemMat, TandemVec??

Almost. Call assembly twice if the pointer to Amat is different to Bmat: Same MatType, different pointer => two different objects (assemble each); same MatType, same pointer => one object used in two places (assemble once)

uphoffc: Sorry, the class has seen some iterations but the name stuck. I agree that it would make sense to rename it in its current state.

PetscBlockVector - same story for the name - is in fact just a wrapper for a petsc Vec, there's nothing more to it. The reason I introduced is that I can keep petsc out of the "src" directory, i.e. I can compile everything in src into a library without dependency on petsc.

Moreover, the assemble function in DGOperator becomes independent of petsc. I could also write an "EigenMatrix" and an "EigenLinearSolver" class and I wouldn't need to change a line in DGOperator.

PR #25, opened 2021-11-16
Enhancement: Add CSV standard comment marker

dmay23:
Marking header lines with a `#` will permit trivial parsing of these output files with standard CSV headers which support white space delimiters.
Note: Existing parsers currently in use may break.

### DIFF
--- a/app/tandem/SEAS.cpp
+++ b/app/tandem/SEAS.cpp
@@ -77,29 +77,33 @@ auto add_writers(Config const& cfg, LocalSimplexMesh<DomainDimension> const& mes
         }
     }
     if (cfg.fault_probe_output) {
+        auto const& oc = *cfg.fault_probe_output;
         monitor.add_writer(std::make_unique<seas::FaultProbeWriter<DomainDimension>>(
-            cfg.fault_probe_output->prefix, cfg.fault_probe_output->probes,
-            cfg.fault_probe_output->make_adaptive_output_interval(), mesh, cl, fault_map, comm));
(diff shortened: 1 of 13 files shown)


uphoffc:
What do you think of adding proper csv output, i.e. such that a user can choose between CSV and Tecplot? Besides breaking the existing scripts, there's also no proper csv header.

dmay23:
I agree that there is no formal standard for CSV format. However there is a wide spread agreement that prior to the column data there may be any number of "comment" or "header" lines. A wide spread convention by CSV parsers is that such lines start with a `#`. Although, many readers allow you to defined another character to indicate such comment / header lines.

We could also change the white space delimiter to be a `,` - at least then it would be a COMMA separated file.

uphoffc:
I made a few changes and added the option to select the output type with type="tecplot" or type="csv" in the fault_probe_output / domain_probe_output / fault_scalar_output sections. The new default is CSV as, as you pointed out rightly, is the more common format.

Example CSV output:
```csv
# Station dp000 (x = [-2.15653e-17, -3.07588e-17])
"Time","state","slip0","traction0","slip-rate0","normal-stress"
0.000000000000000e+00,5.999999999999996e-01,0.000000000000000e+00,-2.654612236050891e+01,1.000000000000001e-09,4.999999999999999e+01
9.498552887595372e+06,7.062034859526083e-01,1.554866115205896e-05,-2.655174217824122e+01,2.469008134901653e-14,4.999999999999999e+01
```

**Python**
```python
import pandas as pd
pd.read_csv('fltst_dp000.csv', comment='#')
```
**R**
```R
d <- read.csv('fltst_dp000.csv', comment.char='#')
```

PR #33, opened 2021-12-02
Added basic functionality required to checkpoing a SEAS model. This c…

dmay23:
Please review.

The idea is very simple. Every time you run TSSolve, a file is emitted `ts_checkpoint.bin`.
Before call to TSSolve, we look for the existence `ts_checkpoint.bin`. If it exists, we load it and start the time integration from the loaded state. If you want to start a new job, make sure you delete `ts_checkpoint.bin`.

In a second pass I will modify `solve()` to checkpoint every N steps.

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -183,13 +183,13 @@ set(APP_COMMON_SRCS
     localoperator/RateAndStateBase.cpp
     #pc/lspoly.c
     pc/register.cpp
+    common/ts_util.c
+    common/vecnest_util.c
 )
 if(${LAPACK_FOUND})
     list(APPEND APP_COMMON_SRCS
          pc/eigdeflate.c
          pc/reig_aux.c
-         common/ts_util.c
-         common/vecnest_util.c
     )
 endif()
 add_library(app-common ${APP_COMMON_SRCS})
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -188,6 +188,8 @@ if(${LAPACK_FOUND})
     list(APPEND APP_COMMON_SRCS
          pc/eigdeflate.c
          pc/reig_aux.c
+         common/ts_util.c
+         common/vecnest_util.c
     )
 endif()
 add_library(app-common ${APP_COMMON_SRCS})
--- a/app/common/ts_util.c
+++ b/app/common/ts_util.c
@@ -2,6 +2,172 @@
 #include <petsc.h>
 #include <petscts.h>
 #include <petsc/private/tsimpl.h>
+#include "vecnest_util.h"
+
+#include <errno.h>
+#include <sys/stat.h>
+#include <limits.h>
+
+/*
+https://stackoverflow.com/questions/2336242/recursive-mkdir-system-call-on-unix
+*/
+static int recursive_mkdir(const char *dir)
+{
+  int    num,error_number;
+  char   tmp[PATH_MAX];
+  char   *p = NULL;
+  size_t len;
+  
+  snprintf(tmp,sizeof(tmp),"%s",dir);
+  len = strlen(tmp);
+  if (tmp[len - 1] == '/') {
+    tmp[len - 1] = 0;
+  }
+  for (p = tmp + 1; *p; p++) {
+    if (*p == '/') {
+      *p = 0;
+      num = mkdir(tmp,S_IRWXU);
+      error_number = errno;
+      if (num != 0) {
+        switch(error_number) {
+          case EEXIST:
+            printf("[%s] directory exists\n",tmp);
+            break;
+          case EACCES:
+            printf("[error %s] See EACCES man mkdir (code %d)\n",tmp,error_number);
+            break;
+          case EBADF:
+            printf("[error %s] See EBADF man mkdir (code %d)\n",tmp,error_number);
+            break;
+          case EDQUOT:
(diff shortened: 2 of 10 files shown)


app/common/ts_util.c
```diff
@@ -0,0 +1,252 @@
...
+  ierr = PetscViewerDestroy(&viewer);CHKERRQ(ierr);
+  /*
+  TSView(ts,PETSC_VIEWER_STDOUT_WORLD);
+  {
+    PetscReal time;
+    PetscInt tk;
+    TSGetStepNumber(ts,&tk); printf("  step %d\n",(int)tk);
+    TSGetTime(ts,&time); printf("  time %lf\n",time);
+  } 
+  */
```
uphoffc, line 221: ?

app/common/ts_util.c
```diff
@@ -0,0 +1,252 @@
+
+#include <petsc.h>
+#include <petscts.h>
+#include <petsc/private/tsimpl.h>
```
uphoffc, line 4: Is this header part of the petsc installation?

app/common/vecnest_util.c
```diff
@@ -0,0 +1,73 @@
+
+#include <petsc.h>
+#include <petscvec.h>
+#include <petsc/private/vecimpl.h>
```
uphoffc, line 4: Is this header part of the petsc installation? (And why is it needed here?)

app/common/ts_util.c
```diff
@@ -0,0 +1,252 @@
...
+    TSGetStepNumber(ts,&tk); printf("  step %d\n",(int)tk);
+    TSGetTime(ts,&time); printf("  time %lf\n",time);
+  } 
+  */
+  PetscFunctionReturn(0);
+}
+
+PetscErrorCode tandem_TSLoad(TS ts)
+{
+  const char filename[] = "ts_checkpoint.bin";
```
uphoffc, line 227: Please make the filename a function argument, then we can add it to the toml config.

uphoffc (changes requested):
Do I understand correctly that the only changes needed in _TSView is
```c
  if (!ts->adapt) {
      ierr = TSAdaptCreate(PetscObjectComm((PetscObject)ts),&ts->adapt);CHKERRQ(ierr);
      ierr = TSAdaptSetType(ts->adapt,TSADAPTNONE);CHKERRQ(ierr);
    }
```
If yes, couldn't this code be moved outside of _TSView, e.g. below TSSetFromOptions or so? Then there wouldn't be the need to duplicate internal PETSc code in this repo until the bug gets fixed.

dmay23:
> Do I understand correctly that the only changes needed in _TSView is
> 
> ```c
>   if (!ts->adapt) {
>       ierr = TSAdaptCreate(PetscObjectComm((PetscObject)ts),&ts->adapt);CHKERRQ(ierr);
>       ierr = TSAdaptSetType(ts->adapt,TSADAPTNONE);CHKERRQ(ierr);
>     }
> ```
> 
> If yes, couldn't this code be moved outside of _TSView, e.g. below TSSetFromOptions or so? Then there wouldn't be the need to duplicate internal PETSc code in this repo until the bug gets fixed.

Unfortunately no. There is also (above it) a bunch of data (like the current time!) which I also had to write into the file which PETSc didn't do (embarassingly). So we cannot get around this huge copy-and-paste until PETSc is fixed.

app/common/vecnest_util.c
dmay23: > 

Yes it is. It is needed as in the last function I need to set set some function points (`x->ops->`). There might be an API available to set those but I don't think so. There is for Mat and PC (for use with the Shell's) but not for Vec (to my knowledge).

app/common/ts_util.c
dmay23: > 

Yes

app/common/ts_util.c
dmay23: > 

Can be deleted. Sanity checking.

app/CMakeLists.txt
```diff
@@ -188,6 +188,8 @@ if(${LAPACK_FOUND})
     list(APPEND APP_COMMON_SRCS
          pc/eigdeflate.c
          pc/reig_aux.c
+         common/ts_util.c
+         common/vecnest_util.c
```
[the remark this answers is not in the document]
dmay23: > 

I don't. Yes these files don't need to be guarded.

app/common/ts_util.c
dmay23: Will do.

dmay23: Done.

dmay23:
I am ready for a second round of reviews. There is a new method `solve_with_checkpoints()` which wraps TSSolve and executes in stages of N steps, where N is the requested checkpoint frequency.

Currently SEAS runs cannot be reliably restarted with the functionality here. What I observe is that upon some restarts, an exception will immediately occur within the friction solver producing an error like 
```
sigma_n = 50
|tau| = 3.31016e+12
psi = inf
L = 0
U = 7.15797e+11
F(L) = -nan
F(U) = -inf
terminate called after throwing an instance of 'tndm::function_nan_inf'
  what():  NaN or Inf detected: F(x) = -nan with x = 0.000000
[kernelpanic:102592] *** Process received signal ***
```

Such behaviour can be reproduced by running bp1 and halting execution during between step 50 and 60 and attempting a restart.

This branch is WIP and should not be merged until it is robust.

PR #44, opened 2023-03-30
Document installation procedure with spack

Thomas-Ulrich:
I've added documentation of the procedure to install tandem with spack.
Note that it will not fully work now, until 
- the tandem module is reviewed and accepted by spack (should take less than 1 week).
- the PR with missing include is merged.
(as a work-around, following the procedure for installing tandem on SuperMUC should work, apart from the include).

### DIFF
--- a/docs/getting-started/installation.rst
+++ b/docs/getting-started/installation.rst
@@ -147,6 +147,7 @@ Install METIS and ParMETIS
     # make config && make && make install
     # cd ../parmetis-4.0.3
     # make config && make && make install
+    # cd ..
 
 Install PETSc
 -------------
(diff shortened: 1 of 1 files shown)


daisy20170101:
Thanks! there are several versions of spack on NG already. Could I use them or I have to install to my own path. 

supermucNG:
spack/21.1.1  spack/22.1.0  spack/22.2.1  spack/release/21.1.1

Thomas-Ulrich:
I use the module (preloaded by default) spack/22.2.1

Thomas-Ulrich:
Note that `spack/22.2.1` is not to load spack, but to load the `spack software stack`, that is many environment modules installed with spack.
The spack module is `user_spack`.

NicoSchlw:
Hi, the @main in line 94 needs to be moved behind tandem

Thomas-Ulrich:
> Hi, the @ main in line 94 needs to be moved behind tandem

Thank you, I've fixed it.

AliceGabriel:
It looks good to me, would be curious on @dmay23 review before merging

PR #55, opened 2023-09-28
Add azure pipelines

Thomas-Ulrich:
Add azure-pipelines 
- azure-pipelines-template.yml build the application and run the test (it basically does the same as the .travis.yml file, but faster). As in seissol, we could test the build with different option, by adding parameters, see https://github.com/SeisSol/SeisSol/blob/master/azure-pipelines.yml

- azure-pipelines-clang-format.yml verify that the src, app, and test are formated accorded to clang-format-11.
(it is copied and adapted from SeisSol)

None of the tests currently pass (https://tulrich.visualstudio.com/thomasulrich/_build/results?buildId=60&view=results), because:
- a few files are not clang-formatted

- the doctest.h is too old
```
[ 55%] Building CXX object test/CMakeFiles/test-runner.dir/main.cpp.o
In file included from /home/vsts/work/1/s/test/main.cpp:2:
/home/vsts/work/1/s/test/../external/doctest.h:4036:47: error: size of array ‘altStackMem’ is not an integral constant-expression
 4036 |         static char             altStackMem[4 * SIGSTKSZ];
      |                                               ^
make[2]: *** [test/CMakeFiles/test-runner.dir/build.make:76: test/CMakeFiles/test-runner.dir/main.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1643: test/CMakeFiles/test-runner.dir/all] Error 2
```

These 2 problems are addressed by separated #56 and #57.
The PR is ready, but, I would suggest first reviewing and merging #56  and #57 into #55, therefore draft.
Note, we also need to register the workflow to visualstudio.com, but I don't have the rights to do so
(therefore I only tested on my fork).


Thomas-Ulrich:
will be implemented using github workflows in #69

PR #56, opened 2023-09-28
update doctest.h

Thomas-Ulrich:
(note that this PR merges into thomas/azure and not in main).
Fixes the build in the Azure pipeline, see e.g.
https://tulrich.visualstudio.com/thomasulrich/_build/results?buildId=61&view=results


Thomas-Ulrich:
not required any more

PR #68, opened 2024-06-20
Compilation bug fix after the documentation update

JeenaYun:
The bug was rising from the incorrect syntax of end of line mark (;) after adding help strings.
The bug is fixed and I confirmed that the compilation works well.


Thomas-Ulrich (approved):
LGTM

PR #69, opened 2024-06-20
Fixes GPU, update for clang and latest Petsc, add workflow  testing build

Thomas-Ulrich:
(building-up on dmay/petsc_dev_hip branch)
- fixes tandem GPU #52  by using the allocating vector consistently with the given vec_type
- enforce MatSetBlockSize
- update code to compile with Clang and latest Petsc
- add gh workflow to test compilation of tandem with gcc 13 and clang 18


hpc4geo (commented):
@Thomas-Ulrich Where does the value of `env.PETSC_VERSION` come from? We will need to frequently bump the version of petsc being used over time.

Thomas-Ulrich:
env.PETSC_VERSION was hardcoded. Now the code looks at the pypi server to find the latest release.
(and then we install the latest release)

PR #78, opened 2024-10-01
Changing the velocity root-finding 

yohaimagen:
Changing the velocity root-finding, line 13 of algorithm 1 in the tandem paper.
Instead of searching for $V$ S.T
$\sigma_n F(V, \psi_{k}^{e}) + \eta V - T = 0$
searching for $\xi$ S.T
$V = 10^{\xi}$

It was tested against bp1 bp3, and my own SSEs simulations and it shows ~40% improvement in run time when running with green functions. 

bp1:
new root-finding run time: 554.237
main run time: 888.816
bp1

bp3:
new root-finding run time: 4099.94
main run time: 6373.47
bp3

Is more testing needed before merging into main? If so, what would you like to test?

### DIFF
--- a/app/localoperator/DieterichRuinaAgeing.h
+++ b/app/localoperator/DieterichRuinaAgeing.h
@@ -80,8 +80,7 @@ public:
 
     auto slip_rate(std::size_t index, double sn,
                    std::array<double, TangentialComponents> const& tau, double psi) const
-        -> std::array<double, TangentialComponents> 
-    {
+        -> std::array<double, TangentialComponents> {
         auto eta = p_[index].get<Eta>();
         auto tauAbsVec = tau + p_[index].get<TauPre>();
         double snAbs = -sn + p_[index].get<SnPre>();
@@ -90,52 +89,40 @@ public:
         double a = -32;
         double a_min = std::log10(std::nextafter(0, INFINITY));
         double b;
-        if (eta == 0.0)
-        {
+        if (eta == 0.0) {
             V = Finv(index, snAbs, tauAbs, psi);
-        }
-        else
-        {
-            if (snAbs <= 0.0)
-            { /* Implies the fault is experiencing tension / opening */
-                snAbs = 0.0; /* Just to illustrate what we are doing */
+        } else {
+            if (snAbs <= 0.0) { /* Implies the fault is experiencing tension / opening */
+                snAbs = 0.0;    /* Just to illustrate what we are doing */
                 /* Solve R(V) = T - sigma_n F(V,psi) - eta V with sigma_n = 0.0 */
                 V = tauAbs / eta;
-            }
-            else
-            {
+            } else {
                 b = std::log10(tauAbs / eta);
-                auto fF = [this, &index, &snAbs, &tauAbs, &psi, &eta](double Ve)
-                {
-                    return tauAbs - this->F(index, snAbs, std::pow(10.0,Ve), psi) - eta * std::pow(10.0,Ve);
+                auto fF = [this, &index, &snAbs, &tauAbs, &psi, &eta](double Ve) {
+                    return tauAbs - this->F(index, snAbs, std::pow(10.0, Ve), psi) -
+                           eta * std::pow(10.0, Ve);
                 };
-                try
-                {
+                try {
                     auto Ve = zeroIn(a, b, fF);
                     V = std::pow(10.0, Ve);
-                }
-                catch (std::exception const&)
-                {
-                    try
-                    {
+                } catch (std::exception const&) {
+                    try {
                         auto Ve = zeroIn(a_min, b, fF);
                         V = std::pow(10.0, Ve);
-                    }
-                    catch(std::exception const&)
-                    {
+                    } catch (std::exception const&) {
                         std::cout << "sigma_n = " << snAbs << std::endl
-                            << "|tau| = " << tauAbs << std::endl
-                            << "psi = " << psi << std::endl
-                            << "L = " << a << std::endl
-                            << "U = " << b << std::endl
-                            << "F(L) = " << fF(a) << std::endl
-                            << "F(U) = " << fF(b) << std::endl;
+                                  << "|tau| = " << tauAbs << std::endl
+                                  << "psi = " << psi << std::endl
+                                  << "L = " << a << std::endl
+                                  << "U = " << b << std::endl
+                                  << "F(L) = " << fF(a) << std::endl
+                                  << "F(U) = " << fF(b) << std::endl;
                         throw;
                     }
                 }
             }
         }
-    return -(V / tauAbs) * tauAbsVec;
+        return -(V / tauAbs) * tauAbsVec;
     }
 
     double state_rhs(std::size_t index, double V, double psi) const {
--- a/app/localoperator/DieterichRuinaAgeing.h
+++ b/app/localoperator/DieterichRuinaAgeing.h
@@ -80,14 +80,16 @@ public:
 
     auto slip_rate(std::size_t index, double sn,
                    std::array<double, TangentialComponents> const& tau, double psi) const
-        -> std::array<double, TangentialComponents> {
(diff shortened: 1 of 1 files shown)


yohaimagen:
@dmay23 
@AliceGabriel

app/localoperator/DieterichRuinaAgeing.h
```diff
@@ -99,19 +99,20 @@ class DieterichRuinaAgeing {
...
                     std::swap(a, b);
                 }
-                auto fF = [this, &index, &snAbs, &tauAbs, &psi, &eta](double V) {
-                    return tauAbs - this->F(index, snAbs, V, psi) - eta * V;
+                auto fF = [this, &index, &snAbs, &tauAbs, &psi, &eta](double Ve) {
+                    return tauAbs - this->F(index, snAbs, std::pow(10.0,Ve), psi) - eta * std::pow(10.0,Ve);
                 };
                 try {
-                    V = zeroIn(a, b, fF);
+                    auto Ve = zeroIn(-216, std::log10(b), fF);
```
Thomas-Ulrich, line 106: What about this hardcoded -216?
Do you mean the smallest double? (1e−308 or so?).
Maybe increasing this number a bit could save you a few iterations? (e.g. -50 or so?).

app/localoperator/DieterichRuinaAgeing.h
```diff
@@ -99,19 +99,20 @@ class DieterichRuinaAgeing {
...
+                    V = std::pow(10.0,Ve);
                 } catch (std::exception const&) {
                     std::cout << "sigma_n = " << snAbs << std::endl
                               << "|tau| = " << tauAbs << std::endl
                               << "psi = " << psi << std::endl
                               << "L = " << a << std::endl
                               << "U = " << b << std::endl
-                              << "F(L) = " << fF(a) << std::endl
-                              << "F(U) = " << fF(b) << std::endl;
+                              << "F(L) = " << fF(-216) << std::endl
```
Thomas-Ulrich, line 114: I would initialize with:
```
a = std::log10(-50);
b=std::log10(tauAbs / eta);
```
and do not change the code here.

app/localoperator/DieterichRuinaAgeing.h
yohaimagen: I removed the hardcoded -216, which was somewhat of a random choice after -32 did not work. Now, I implemented a loop to apply a = -2^-5 ... -2^-8, which saves some more time (I'm currently running the benchmark test).

app/localoperator/DieterichRuinaAgeing.h
yohaimagen: Done.

app/localoperator/DieterichRuinaAgeing.h
```diff
@@ -86,37 +86,54 @@ class DieterichRuinaAgeing {
         double snAbs = -sn + p_[index].get<SnPre>();
         double tauAbs = norm(tauAbsVec);
         double V = 0.0;
-        if (eta == 0.0) {
+        double max_xi = -std::pow(2, 8);
```
hpc4geo, line 89: For consistency with code you added further down, it would make sense that this is
```
int max_xi = 8
```

app/localoperator/DieterichRuinaAgeing.h
```diff
@@ -86,37 +86,54 @@ class DieterichRuinaAgeing {
...
-                    std::cout << "sigma_n = " << snAbs << std::endl
-                              << "|tau| = " << tauAbs << std::endl
-                              << "psi = " << psi << std::endl
-                              << "L = " << a << std::endl
-                              << "U = " << b << std::endl
-                              << "F(L) = " << fF(a) << std::endl
-                              << "F(U) = " << fF(b) << std::endl;
-                    throw;
-                }
+                for (int xi = 5; xi < 9; ++xi)
```
hpc4geo, line 110: This would become
```
xi <= max_xi
```

app/localoperator/DieterichRuinaAgeing.h
```diff
@@ -86,37 +86,54 @@ class DieterichRuinaAgeing {
...
+                    try
+                    {
+                        a = -std::pow(2.0, xi);
+                        auto Ve = zeroIn(a, b, fF);
+                        V = std::pow(10.0, Ve);
+                        break;
+                    }
+                    catch (std::exception const&)
+                    {
+                        if (a == max_xi)
```
hpc4geo, line 121: Equality comparisons between doubles / floats are almost never going to be true. Don't rely on them. I would use (if you adopt changes above)

```
if (xi == max_xi) {
```

hpc4geo (commented):
1) I don't think the loop is a good way to go. It potentially introduces additional function evals of F(), and some are redundant. For example each iteration evals F(a) and F(b). Only a changes, so F(b) is re-computed. The whole point of this PR is to reduce calls to F(). 

2) After adding the loop, the profiling report provided in the PR needs to be updated because the numbers likely change.

3) I am more inclined to remove the loops, go back to a hard coded number. The hard coded number could be defined using `std::nextafter(0, INFINITY)`. `nextafter()` gives you the smallest representable number from 0.0 (in this case).

```
#include <iostream>
#include <cmath>

int main() {
    double x = 0.0;
    double nextSmallest = std::nextafter(x, INFINITY);

    std::cout << "x: " << x << std::endl;
    std::cout << "Next smallest double: " << nextSmallest << std::endl;
    printf("%1.20e\n",nextSmallest);
    return 0;
}
```

yohaimagen:
1. In the vast majority of cases, we are performing just the first iteration of the for loop with `xi=5`. This saves time in the root-finding (searching on shorter intervals).

2. The new profiling report was running during the night and done more robustly, with a comparison of 10 runs of the new root-finding and the main branch version.

bp1:

New root-finding run time: 371.4818

Main run time: 757.6242

bp3:

New root-finding run time: 2179.287

Main run time: 4531.09

Overall, 51% and 52% for bp1 and bp3, respectively, improvement over the main branch and ~10% improvement over the hardcoded very small number version.

3. An alternative approach to the for loop will be to try `xi=5`, and if this fails, to go immediately to `std::nextafter(0, INFINITY)`.

app/localoperator/DieterichRuinaAgeing.h
yohaimagen: Done

app/localoperator/DieterichRuinaAgeing.h
yohaimagen: done

app/localoperator/DieterichRuinaAgeing.h
yohaimagen: done

hpc4geo:
@yohaimagen Could you please implement your suggested option 3?  I really like this idea.

app/localoperator/DieterichRuinaAgeing.h
```diff
@@ -80,43 +80,62 @@ class DieterichRuinaAgeing {
...
         auto eta = p_[index].get<Eta>();
         auto tauAbsVec = tau + p_[index].get<TauPre>();
         double snAbs = -sn + p_[index].get<SnPre>();
         double tauAbs = norm(tauAbsVec);
         double V = 0.0;
-        if (eta == 0.0) {
+        double a = -32;
+        double a_min = std::log10(std::nextafter(0, INFINITY));
+        double b;
+        if (eta == 0.0)
```
Thomas-Ulrich, line 93: [minor suggestion]
to adhere the format of the initial code, you can use clang-format on your file:
`clang-format -i app/localoperator/DieterichRuinaAgeing.h`

dmay23:
For what it is worth, I re-run clang-format over all the files changed in
each branch before I merge them. If people do this themselves this does
reduce work on my side.

app/localoperator/DieterichRuinaAgeing.h
yohaimagen: Done

PR #85, opened 2024-11-05
implement polymorphism-based implementation of the RS friction laws

Thomas-Ulrich:
merges into Jeena's branch implementing the slip law.

state_rhs a pure virtual function in the abstract class DieterichRuinaBase, which is implemented in DieterichRuinaAging and DieterichRuinaSlip.


JeenaYun:
Looks good and compiling well. Thanks, Thomas.

PR #95, opened 2025-07-07
CI pipeline setup and testing

piyushkarki:
# Modify CI Workflows and Add Regression Testing

## Summary

This pull request introduces major updates to the CI system and testing infrastructure for tandem. The key changes include:

* Reusable GitHub Actions workflows
* Docker-based build and test environment
* Regression and consistency testing
* Infrastructure for SEAS cycle simulations
* Serves as a foundation to add more tests in the future

---

### CI Workflows

**Added:**
* `.github/workflows/build-and-test-tandem.yml` — Builds and tests across multiple compilers and dimensions
* `.github/workflows/build-dependencies.yml` — Builds the Docker image with project dependencies
* `.github/workflows/check-petsc-version.yml` — Checks for PETSc version updates and triggers dependency rebuilds
* `.github/workflows/main-workflow.yml` — Main entry point that triggers all the workflows in the CI pipeline

**Modified:**
* Renamed `clang-format.yml` to `check-clang-format.yml` for clarity

**Removed:**
* Deprecated `build-tandem.yml` workflow

### Docker Support

* Introduced a new `Dockerfile` that:
  * Installs all necessary system and Python dependencies
  * Builds PETSc with a specified version (major=3, minor=22, patch=latest available)
  * Sets up a non-root user for MPI compatibility
  * Creates a Python virtual environment for test tooling

### Testing Infrastructure

* Added comprehensive test suite:
  * Static regression tests (2D and 3D)
  * SEAS regression tests (2D)
  * Parallel consistency checks (2D and 3D)
  * Convergence tests (2D and 3D)
* Added test utilities in `conftest.py`:
  * Fixtures for loading VTU files
  * Output comparison logic
  * Error computations
  * Tolerance values to be decided and modified
* Integrated test data as a separate repo pulled later

---

## Testing

The CI pipeline performs the following:

1. Checks code formatting with `clang-format`
2. Verifies PETSc version compatibility
3. Builds Docker-based dependencies if needed
4. Builds and tests the code using:
   * Multiple compilers (GCC 13, Clang 18)
   * 2D and 3D configurations
   * Polynomial degree 2 & 3
5. Runs the full test suite (`make test`)

---

## Test Suite Details

The test suite is divided into **C++ Unit Tests** for fine-grained component checks and **Pytest Integration Tests** for overall simulation validation.

### Test Directory Structure

Test configurations and generator scripts are maintained directly within the repository to keep tests closely coupled with the code.

- **`reference_configs/`**: Contains simulation configurations (`.toml`, `.lua`, `.geo`) separated by dimension (`2D/`, `3D/`).
- **`scripts/`**: Contains the individual shell scripts used to generate reference output data for each test.
- **`test_data/`**: The destination folder where the external `tandem_test_data` repository is cloned (only when integration tests are enabled).

#### Basic Unit Tests

By default, compiling the tests will only prepare the local C++ unit tests. Run them using:

```bash
make test
```

#### Integration Tests (CI)

Integration tests require heavy reference data (meshes, Green's functions, `.vtu` files). Before running them, install the required Python dependencies:

```bash
pip install -r test/requirements.txt
```

Then compile with:

```bash
cmake -DRUN_CI_TESTS=ON ..
```

> **Note:** This flag automatically clones the `tandem_test_data` repository into the local `test_data/` directory.

#### Testing with Custom Data Branches

By default, the CI setup pulls the `master` branch of the `tandem_test_data` repository. If you are developing a new feature and need a custom branch of the test data, you can specify it during configuration using `-DTEST_DATA_BRANCH`.

**Example:** If you are working on a Tandem branch named `<user-name>/add-feature-1`, create a corresponding data branch `<user-name>/add-data-for-feature-1` in the data repo and compile with:

```bash
cmake -DRUN_CI_TESTS=ON -DTEST_DATA_BRANCH=<user-name>/add-data-for-feature-1 ..
```

For more test details, please see the README.md inside the test directory.

## Notes

* Docker image ensures reproducibility and local/CI parity
* This infrastructure is designed for extensibility and future enhancements


.gitmodules
```diff
@@ -4,3 +4,7 @@
 [submodule "submodules/yateto"]
 	path = submodules/yateto
 	url = https://github.com/SeisSol/yateto.git
+[submodule "test/test_data"]
```
Thomas-Ulrich: Do you really want to introduce a (probably heavy and for most users unecessary) submodule?

.gitignore
```diff
@@ -12,3 +12,6 @@ docs/_build/
 
 # Mesh files
 examples/**/*.msh
+
+# Test output foldersd ..
```
Thomas-Ulrich: typo

test/test_convergence_static.py
```diff
@@ -0,0 +1,79 @@
...
+                    data.append(
+                        (
+                            current_entry["h"],
+                            current_entry["dofs"],
+                            current_entry["residual"],
+                            current_entry["l2_error"],
+                            current_entry["h1_error"],
+                        )
+                    )
+                    current_entry = {}
```
Thomas-Ulrich, line 40: If a log entry is missing the H1-semi error line (perhaps due to a simulation crash), the next simulation's data will be merged into the old dictionary, potentially creating "ghost" data.

So maybe:
```python 
            if line.startswith("DOFs:"):
                current_entry = {}
                current_entry["dofs"] = int(line.split()[1])
```

test/test_static_convergence.py
```diff
@@ -0,0 +1,79 @@
...
+                            current_entry["h"],
+                            current_entry["dofs"],
+                            current_entry["residual"],
+                            current_entry["l2_error"],
+                            current_entry["h1_error"],
+                        )
+                    )
+                    current_entry = {}
+
+    data.sort(key=lambda x: x[0])  # sort by mesh size
```
Thomas-Ulrich: `data.sort(key=lambda x: x["h"])`?

test/test_static_convergence.py
```diff
@@ -0,0 +1,79 @@
...
+
+def compute_all_slopes(data):
+    mesh_size = np.array([entry[0] for entry in data])
+    l2_errors = np.array([entry[3] for entry in data])
+    h1_errors = np.array([entry[4] for entry in data])
+
+    slope_l2 = compute_slope(mesh_size, l2_errors)
+    slope_h1 = compute_slope(mesh_size, h1_errors)
+
+    # return slope_l2, slope_h1
```
Thomas-Ulrich: uncomment !

test/test_regression_SEAS.py
```diff
@@ -0,0 +1,106 @@
...
+    i = 0
+    n = len(time)
+
+    while i < n:
+        window_start = time[i]
+        window_end = window_start + window_size
+
+        # Find the indices within the window
+        window_mask = (time >= window_start) & (time <= window_end)
+        window_indices = np.where(window_mask)[0]
```
Thomas-Ulrich: Inside a while loop, you are scanning the entire time array for every single iteration. If your CSV has 1,000,000 rows, this will be incredibly slow.

you could use instead:
```
 window_indices = np.searchsorted(time, t_end, side='right')
```

test/test_regression_SEAS.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if len(window_indices) < 3:
```
Thomas-Ulrich: break?

test/test_regression_SEAS.py
```diff
@@ -0,0 +1,106 @@
...
+
+            # Move to end of window to avoid overlapping detections
+            i = window_indices[-1] + 1
+        else:
+            i += 1
+    return events
+
+
+def check_SEAS_consistency(file_vmax, file_vmax_gf, tolerances):
+    window_size = 1e9  # seconds
```
Thomas-Ulrich: This will only detect one event per 31 years. Intended?

test/test_regression_SEAS.py
```diff
@@ -0,0 +1,106 @@
...
+    arr2 = np.array(sorted_events_gf)
+
+    times1 = arr1[:, 0]
+    times2 = arr2[:, 0]
+
+    values1 = arr1[:, 1]
+    values2 = arr2[:, 1]
+
+    assert np.allclose(
+        times1, times2, rtol=tolerances["seas"]
```
Thomas-Ulrich: this will break if the number of events is not the same.

test/test_regression_SEAS.py
```diff
@@ -0,0 +1,106 @@
...
+
+    event_time_interval_QD = [
+        events[i + 1][0] - events[i][0] for i in range(len(events) - 1)
+    ]
+    event_time_intervals_QDGreen = [
+        events_gf[i + 1][0] - events_gf[i][0] for i in range(len(events_gf) - 1)
+    ]
+    assert np.allclose(
+        event_time_interval_QD,
+        event_time_intervals_QDGreen,
```
Thomas-Ulrich: I would remove all references to QD and QDGreen in this function.
Change to reference and tested or something like that.

test/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+import pytest
+from pathlib import Path
+
+
+@pytest.fixture(scope="module")
+def tolerances():
+    return {
+        "static": 1e-8,
+        "seas": 1e-8,
+        "seas_events": 1e-2,
```
Thomas-Ulrich: so high?

test/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+
+        field_components = field_array.GetNumberOfComponents()
+        n_cells = data.GetNumberOfCells()
+
+        point_to_value_mapping = {}
+        for cell_id in range(n_cells):
+            cell = data.GetCell(cell_id)
+            point_ids = [cell.GetPointId(i) for i in range(cell.GetNumberOfPoints())]
+            coords = [data.GetPoint(pid) for pid in point_ids]
+            dimensions = 3
```
Thomas-Ulrich: hardcoded dimension intended?

test/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+        reader.SetFileName(file_path)
+        reader.Update()
+        return reader.GetOutput()
+
+    return _load
+
+
+@pytest.fixture(scope="module")
+def get_cell_centroid_point_dofs():
+    def _get(data, field_name):
```
Thomas-Ulrich, line 160: Might be useful to put in the docstring an example of format for point_to_value_mapping, e.g.

{
    (0.5, 0.5, 0.0): {           # <-- The Cell Centroid (Key)
        (0.0, 0.0, 0.0): 101.3,  # Corner 1: (coords): value
        (1.0, 0.0, 0.0): 101.5,  # Corner 2: (coords): value
        (1.0, 1.0, 0.0): 102.1,  # Corner 3: (coords): value
    }
}

Actually probably Quadrature point, right?

test/test_parallel_consistency_static.py
```diff
@@ -0,0 +1,52 @@
...
+    compute_l2_error_with_reference_data,
+    tolerances,
+):
+    ref_data = load_vtu_file(reference_file)
+    output_procs = [2**j for j in range(1, number_of_configs)]
+    vtu_files = [f"{file_prefix}{i}_0.vtu" for i in output_procs]
+
+    ref_dict = get_cell_centroid_point_dofs(ref_data, field_name)
+    for idx, vtu_file in enumerate(vtu_files):
+        data = load_vtu_file(vtu_file)
```
Thomas-Ulrich, line 23: simpler to use pyvista?

test/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@pytest.fixture(scope="module")
+def compute_l2_error_with_reference_data():
+    def _compute(ref_dict, cmp_dict):
+        coords = []
+        errors = []
+        for centroid, points in ref_dict.items():
+            if centroid not in cmp_dict:
+                print(f"Centroid {centroid} not found in comparison data")
```
Thomas-Ulrich: Doesn't it happen sometimes that the centroid differs by eps = the double precision?

test/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+            for point, ref_dof in points.items():
+                if point not in cmp_points:
+                    print(
+                        f"Point {point} not found in comparison data for centroid {centroid}"
+                    )
+                    continue
+
+                cmp_dof = cmp_points[point]
+                abs_diff = abs(cmp_dof - ref_dof)
+                coords.append(point)
```
Thomas-Ulrich: coords is not used, right?

test/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+        vtu_files = [f"{file_prefix}{i}_0.vtu" for i in output_procs]
+
+        ref_dict = get_cell_centroid_point_dofs(ref_data, field_name)
+        for idx, vtu_file in enumerate(vtu_files):
+            data = load_vtu_file(vtu_file)
+            cmp_dict = get_cell_centroid_point_dofs(data, field_name)
+            L2_norm = compute_l2_error_with_reference_data(ref_dict, cmp_dict)
+            assert (
+                L2_norm < tolerances["static"]
+            ), f"Output for {output_procs[idx]} procs is inconsistent with output for 1 proc."
```
Thomas-Ulrich: print reference_file and file_prefix in the log?

test/conftest.py
```diff
@@ -0,0 +1,143 @@
...
+                coords.append(point)
+                errors.append(abs_diff)
+        L2_norm = np.sqrt(np.sum(np.square(errors)))
+        return L2_norm
+
+    return _compute
+
+
+@pytest.fixture(scope="module")
+def compare_one_vs_many_processor_outputs(
```
Thomas-Ulrich: It seems this function is not used (or duplicated with ./test/test_parallel_consistency.py)

test/test_parallel_consistency.py
```diff
@@ -0,0 +1,52 @@
...
+    temp_results_path,
+    load_vtu_file,
+    get_cell_centroid_point_dofs,
+    compute_l2_error_with_reference_data,
+    tolerances,
+):
+    domain_dimension = request.config.getoption("domain_dimension")
+    reference_file = temp_results_path / f"parallel_output{domain_dimension}D_1_0.vtu"
+    output_prefix = temp_results_path / f"parallel_output{domain_dimension}D_"
+    field_name = "u0"
```
Thomas-Ulrich: only u0? Can't we compare all fields?

test/CMakeLists.txt
```diff
@@ -54,3 +54,38 @@ doctest_discover_tests(test-tensor)
 add_executable(test-util util.cpp)
 target_link_libraries(test-util test-runner)
 doctest_discover_tests(test-util)
+
+if(DEFINED RUN_CI_TESTS)
+    find_package(Python COMPONENTS Interpreter Development REQUIRED)
+
+    add_test(NAME generate-test-outputs
+        COMMAND bash -c
```
Thomas-Ulrich, line 104: Hardcoding bash -c makes this build file fail on Windows.

test/CMakeLists.txt
```diff
@@ -54,3 +54,38 @@ doctest_discover_tests(test-tensor)
...
+        "source ${CMAKE_CURRENT_SOURCE_DIR}/test_data/scripts/generate_test_outputs.sh ${DOMAIN_DIMENSION} ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}"
```
Thomas-Ulrich: You probably need to consider race conditions if CTest runs tests in parallel.
(use set_tests_properties to define a dependency chain)

test/CMakeLists.txt
```diff
@@ -54,3 +54,38 @@ doctest_discover_tests(test-tensor)
 add_executable(test-util util.cpp)
 target_link_libraries(test-util test-runner)
 doctest_discover_tests(test-util)
+
+if(DEFINED RUN_CI_TESTS)
```
Thomas-Ulrich: Missing FIXTURES (ask gemini)

.github/workflows/build-and-test-tandem.yml
```diff
@@ -0,0 +1,82 @@
...
+                      -DPOLYNOMIAL_DEGREE=${{ matrix.degree }} \
+                      -DCMAKE_C_COMPILER=\$CC \
+                      -DCMAKE_CXX_COMPILER=\$CXX \
+                      -DRUN_CI_TESTS=ON
+              make -j$(nproc)"
+
+      - name: Run Tests
+        run: |
+          IMAGE=${{ inputs.dependencies_image }}:${{ inputs.petsc_version }}
+          docker run --rm \
```
Thomas-Ulrich, line 63: this second run does not have 
```
            -e CC=${{ matrix.compiler.cc }} \
            -e CXX=${{ matrix.compiler.cxx }} \
```
(cmake might lose track of which compiler it was using.)

.github/workflows/build-and-test-tandem.yml
```diff
@@ -0,0 +1,82 @@
...
+            -v ${{ github.workspace }}:/app \
+            -w /app \
+            -e CC=${{ matrix.compiler.cc }} \
+            -e CXX=${{ matrix.compiler.cxx }} \
+            --user tandem \
+            $IMAGE \
+            bash -c "
+              set -e
+              sudo update-alternatives --install /usr/bin/cc cc /usr/bin/\$CC 100
```
Thomas-Ulrich: really needed?

.github/workflows/build-dependencies.yml
```diff
@@ -0,0 +1,57 @@
...
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: Dockerfile
+          push: true
+          tags: ${{ inputs.dependencies_image }}:${{ inputs.petsc_version }}
+          platforms: linux/amd64
+          build-args: |
+            PETSC_VERSION=${{ inputs.petsc_version }}
+            CC=gcc-13
```
Thomas-Ulrich, line 61: Possible to update to
gcc-15 and clang-22?

Thomas-Ulrich (commented):
Nice! Have a look at my comments.

.github/workflows/build-and-test-tandem.yml
piyushkarki: Removed redundancy.

.github/workflows/build-and-test-tandem.yml
piyushkarki: The matrix variables should be fixed and the make test doesn't really recompile but just runs tests. I could add them back for readability but would ask you to reconsider.

.github/workflows/build-dependencies.yml
piyushkarki: Supposedly but we would need to add a separate package manager like (PPA) and a change in that might break our code as compared to using the native gcc and clang versions that come out of the box with the Ubuntu version that we are using.

test/CMakeLists.txt
piyushkarki: Added fixtures to avoid out of order testing.

test/CMakeLists.txt
piyushkarki: I assumed tandem is either used on unix/linux based systems or even in windows with WSL. Are there users who use it in complete windows OS settings?

test/CMakeLists.txt
piyushkarki: Added with fixtures.

test/conftest.py
piyushkarki: I assume you're referring to the ```seas_events``` tolerance value. I observe that for different systems it varies between systems (for heisenbug/kernelpanic/memoryleak etc.) and therefore if someone were to run it locally, they would want to see it pass even for their local system. Therefore, it is high at the moment. Otherwise, we could use the value that CI calculates and set the limit to something very low.

test/conftest.py
piyushkarki: Done added it to the doc strings. Actually it's the vertices of the triangle written into the vtu so it should be the basis points that land on the vertices.

test/conftest.py
piyushkarki: Yes because coordinates ultimately use 3 dimensions(x,y,z) for representation in vtk even if there is 2D data.

test/conftest.py
piyushkarki: Did not consider this before. Thanks! Added logic to check centroid but only throwing warning.

test/conftest.py
piyushkarki: True. Artifact from a reader I wrote. Removing now.

test/conftest.py
piyushkarki: Removed unused function.

test/conftest.py
piyushkarki: removed the function completely

test/test_parallel_consistency_static.py
piyushkarki: I found this easier but please add a pyvista version if it is more readable.

test/test_parallel_consistency.py
piyushkarki: Yes. I just added this assuming displacement comparison would be the most crucial and also trying to make the test shorter.

test/test_regression_SEAS.py
piyushkarki: Thanks, this apparently uses binary search underneath so it's definitely better!

test/test_regression_SEAS.py
piyushkarki: The logic should work with continue because while is conditioned for i<n but using break for the newer function.

test/test_regression_SEAS.py
piyushkarki: Yes this was just a way for me to fix the test to bp1 so a window would consist of 1 unique event. And after multiple values this was reasonable to define the window size.

test/test_regression_SEAS.py
piyushkarki: Yes that is also desired. We don't want two different number of events.  But I will add a separate assert for this too. Thanks!

test/test_regression_SEAS.py
piyushkarki: Makes sense. Changing the names to reference and tested.

test/test_convergence_static.py
piyushkarki: Good catch! Thanks! Using this fix.

test/test_static_convergence.py
piyushkarki: ```data``` is a tuple so x[0] would be the correct way to access the first elem of each tuple element. Or are you suggesting to format it in a way that it's more readable?

test/test_static_convergence.py
piyushkarki: Uncommented return.

.gitignore
piyushkarki: Fixed. Thanks!

.gitmodules
piyushkarki: This was necessary because a feature that we also wanted was to be able to run all tests locally if desired. We could change the installation instructions to only recursively update yateto and mneme and only update the test_data submodule if they wish to test locally. Please let me know if there are better alternatives.

Thomas-Ulrich: you can do something like:

https://github.com/SeisSol/easi/blob/master/CMakeLists.txt#L80-L90
That is, conditionally cloning the repository in cmake if running the tests.

.github/workflows/build-dependencies.yml
Thomas-Ulrich: right let's keep them until they update ubuntu.

test/conftest.py
Thomas-Ulrich: It's unexpected, right?
You should get similar results independent of the platform.
Could you open an issue? Sounds like a bug.

test/conftest.py
Thomas-Ulrich: ok, but I guess if with a high order mesh there would be more than 3 points (so I would not call them vertices).

test/test_parallel_consistency_static.py
Thomas-Ulrich: ok, as you wish, 
(in pyvista, you can do things like 
centroids = grid.cell_centers().points)

test/test_parallel_consistency.py
Thomas-Ulrich: Would you mind testing if it make the test really slower?
In SeisSol they are comparing all fields, and sometimes, a discrepancy can be found only on one variable, so I think it would be important to try.

test/test_regression_SEAS.py
Thomas-Ulrich: Maybe @JeenaYun, who has more experience with these things, could decide if it makes sense to look for an event with such a large window.

test/test_static_convergence.py
Thomas-Ulrich: I was just suggesting it for readability.

.gitignore
```diff
@@ -12,3 +12,6 @@ docs/_build/
 
 # Mesh files
 examples/**/*.msh
+
+# Test output folders
+test/test_data/temp_test_results/*
```
Thomas-Ulrich: By the way, do you consider automatically cleaning the temp_test_results before running a new test suite?

Thomas-Ulrich (commented):
Thanks for addressing my comments. See answers below.

yohaimagen:
Hi @piyushkarki would you be able to move the Dockerfile under /docker/tests/ to align with #96 where I placed the release Docker files under  /docker/release/.

test/test_regression_SEAS.py
piyushkarki: True, this is only a naive test I set up. I chose the window in a way that we have only one event within a window and this is specific to the bp1 case we are simulating with the specific files. And also, there are only 3 events that happen for the entire test so I would say it's pretty hard coded and not meant for generality.

test/conftest.py
piyushkarki: Reworded the description.

test/conftest.py
piyushkarki: Yes I think it has to do with the difference in PETSc patch file. I tested with different systems but same PETSc versions and it seems to give identical results. It's the PETSc version diff that causes this since I'm running on PETSc 3.22.0 but we pull the latest patch for the CI. I will report this on an issue after confirmation.

.gitmodules
piyushkarki: Thanks this is much better. Adopted this!

test/test_static_convergence.py
piyushkarki: Done. Made it slightly more readable with named tuples.

.gitignore
piyushkarki: That was not there but good point if someone develops inside docker or runs tests locally. Added this into the CMake. Thanks!

piyushkarki:
Hi @yohaimagen , done. Idk why I can't reply to your comment but I moved the docker file into docker/test/ to align with PR #96 .

piyushkarki:
https://github.com/TEAR-ERC/tandem/pull/97#issuecomment-4120409015

> Hey @piyushkarki, one small request regarding the tandem-base image.
> 
> Currently the image is only tagged with the PETSc version (e.g. ghcr.io/tear-erc/tandem/tandem-base:3.22.5). Could you also push a :latest tag pointing to the same image whenever the base is built?
> That way, the release image workflow in https://github.com/TEAR-ERC/tandem/pull/96 can always reference ghcr.io/tear-erc/tandem/tandem-base:latest without hardcoding a specific PETSc version.
> 
> In build-dependencies.yml, it would just be adding latest alongside the versioned tag in the tags field of the Build and push base image step:
> 
> tags: |
> ${{ inputs.base_image }}:${{ inputs.petsc_version }}
> ${{ inputs.base_image }}:latest

Hi @yohaimagen , moving this here to avoid discussing CI on the mesh parser PR. As you suggested, I also added the "latest" tag version to the image here. Also, I would suggest rebasing your current branch on this one to align with the CI and avoid any surprises for later. Or maybe just check the diff to see if you have everything you need on both branches. If not, please just let me know if any changes are needed. Thank you!

PR #96, opened 2026-03-11
Ymagen/docker release images

yohaimagen:
# Add Docker release images and CI workflow

## Summary

This PR adds pre-built Docker release images for Tandem, published to GitHub Container Registry (GHCR). Users can now run any Tandem configuration without compiling from source.

---

### Docker infrastructure (`docker/release/`)

Three Dockerfiles implement a two-stage (builder + slim runtime) pipeline:

| File | Purpose |
|------|---------|
| `Dockerfile.base` | Builds the shared base image: Ubuntu 22.04 + OpenMPI + PETSc + Eigen + Tandem source |
| `Dockerfile.release` | Builds a single configuration (parameterised by `DIM` and `DEG`) on top of the base |
| `Dockerfile.all` | Builds all 12 configurations (2D/3D × degrees 1–6) in one image and installs named wrapper scripts (`tandem_2d_6p_tandem`, `tandem_3d_3p_tandem`, …) |

All images:
- Target `linux/amd64` only
- Run as a non-root `tandem` user
- Bundle example input files at `/work/examples/`

### CI workflow (`.github/workflows/release-images.yml`)

Three-job GitHub Actions pipeline:

1. **`build-base`** — builds and pushes `ghcr.io/tear-erc/tandem-base` with a run-specific tag to avoid `:latest` races.
2. **`build-images`** — matrix job (2 dims × 6 degrees = 12 parallel builds) that produces individual images `ghcr.io/tear-erc/tandem:<dim>d-<deg>p`.
3. **`build-all`** — builds the all-in-one image `ghcr.io/tear-erc/tandem:all`.

**Triggers:**
- Push to this branch or changes to `docker/release/**` / the workflow file itself (for development/testing)
- Git tags matching `v*` (versioned release) → also pushes `:2d-6p`, `:2d-6p-1.2.0`, `:all`, `:all-1.2.0` tags

### Documentation (`docs/getting-started/`)

- **New page** `docker-release.rst` — covers prerequisites (Docker + host-side Gmsh), the all-in-one image, versioning strategy, volume mounting (`-v $(pwd):/host` to avoid shadowing `/work/examples`), MPI usage, and a step-by-step example workflow (copy `.geo` + `.toml` out → generate mesh with Gmsh on host → run inside container).
- **Removed** `quick-start.rst` — the old development Docker workflow (compile-from-source inside a dev container) is no longer the recommended entry point.
- **Updated** `getting-started/index.rst` — toctree and intro text updated to point users to the release images first, then the installation guide for custom builds.
- **Updated** `docs/index.rst` — added a one-line GHCR mention on the main landing page.

---

## Tagging strategy

| Trigger | Tags pushed |
|---------|------------|
| Branch push | `:all-<branch-name>`, `:2d-6p-<branch-name>` (testing only) |
| `v*` git tag | `:all`, `:all-latest`, `:all-1.2.0`, `:2d-6p`, `:2d-6p-latest`, `:2d-6p-1.2.0` |

---

@dmay23 I do not have the permission to make those images public  
The GHCR package ghcr.io/tear-erc/tandem is currently private — this is GitHub's default for newly created packages
To make the images publicly pullable without authentication, an org admin needs to change the visibility once:

Go to https://github.com/orgs/TEAR-ERC/packages/container/tandem/settings
scroll to Danger Zone → Change visibility → Public
This is a one-time action — all future tags pushed to these packages will automatically inherit public visibility.


yohaimagen:
Taging @dmay23, @hpc4geo for a review.
Thanks

docs/getting-started/docker-release.rst
```diff
@@ -0,0 +1,133 @@
...
+
+   On Apple Silicon Macs (M1/M2/M3), the images run under ``linux/amd64``
+   emulation. This works for testing and small runs, but is significantly
+   slower than native. It is not recommended for large production simulations.
+
+Prerequisites
+-------------
+
+- `Docker Desktop <https://docs.docker.com/get-docker/>`_ installed and running.
+- `Gmsh <https://gmsh.info/>`_ installed on your **host machine** for mesh generation.
```
Thomas-Ulrich: Woud yet be as simple as:
https://github.com/SeisSol/Training/blob/main/Dockerfile#L124

docs/getting-started/docker-release.rst
```diff
@@ -0,0 +1,133 @@
...
+      gmsh -2 bp1_sym.geo
+
+   This produces ``bp1_sym.msh`` in the same directory.
+
+3. **Run Tandem inside the container**:
+
+   .. code-block:: sh
+
+      # Inside the container:
+      mpirun -n 4 tandem_2d_3p_tandem /host/bp1_sym.toml
```
Thomas-Ulrich: As you are presenting it, the input and output files are written in the container?
That's seems suboptimal, as the output maybe very large.
It should be possible to use the executable from the container with input/outputs outside the container, e.g.:
https://github.com/SeisSol/Training/blob/main/tool.sh#L2

docker/release/Dockerfile.base
Thomas-Ulrich, line 1: Is it not possible to use this dockerfile  
https://github.com/TEAR-ERC/tandem/blob/piyush/thomas/docker/docker/test/Dockerfile
from
https://github.com/TEAR-ERC/tandem/pulls

Or ar least harmonize. I bet they are very similar.
@piyushkarki

docker/release/Dockerfile.all
```diff
@@ -0,0 +1,85 @@
+# syntax=docker/dockerfile:1
+# Build the all-in-one image containing all 12 Tandem builds:
+#   docker buildx build --platform linux/amd64 -f Dockerfile.all -t tandem:all --load .
+
+# ── Stage 1: Build all 12 Tandem configurations from base ────────────────────
+ARG BASE_IMAGE=ghcr.io/tear-erc/tandem-base:latest
+FROM --platform=linux/amd64 ${BASE_IMAGE} AS builder
+
+RUN for DIM in 2 3; do \
```
Thomas-Ulrich: There is probably ways to speed up the compilation with ninja, e.g.
(@piyushkarki probably also true for your docker too).
And same for the dependencies.
See e.g. how we use it with seissol: 
https://seissol.readthedocs.io/en/latest/lumi.html


```
RUN for DIM in 2 3; do \
        for DEG in 1 2 3 4 5 6; do \
            BUILD_DIR="/opt/tandem/build_${DIM}d_${DEG}p" \
            && mkdir -p ${BUILD_DIR} && cd ${BUILD_DIR} \
            && cmake -G Ninja /opt/tandem/tandem \
                -DCMAKE_CXX_COMPILER=mpicxx \
                -DCMAKE_C_COMPILER=mpicc \
                -DCMAKE_PREFIX_PATH="${PETSC_INSTALL};${EIGEN_INSTALL}" \
                -DDOMAIN_DIMENSION=${DIM} \
                -DPOLYNOMIAL_DEGREE=${DEG} \
            && ninja; \
        done; \
    done
```

docker/release/build_all.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Step 1: Build the base image once (PETSc + Eigen + Tandem source)
+echo "==> Building base image..."
+docker buildx build --platform linux/amd64 \
+    -f ${SCRIPT_DIR}/Dockerfile.base \
+    -t tandem-base:latest \
+    --load ${SCRIPT_DIR}
+
+# Step 2: Build all 12 individual images
+for DIM in 2 3; do
```
Thomas-Ulrich: So you end up with 14 dockers? It is not an overkill?
Also each version is compiled twice then?

.github/workflows/release-images.yml
```diff
@@ -0,0 +1,190 @@
+name: Build and Push Tandem Release Images
+
+on:
+  push:
+    branches:
+      - ymagen/docker-release-images    # your PR branch - builds for testing
```
Thomas-Ulrich: need to remove

.github/workflows/release-images.yml
```diff
@@ -0,0 +1,190 @@
...
+  build-images:
+    name: Build tandem:${{ matrix.dim }}d-${{ matrix.deg }}p
+    needs: build-base
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    strategy:
+      fail-fast: false
```
Thomas-Ulrich: probably you want to activate fail-fast, right?

Thomas-Ulrich (changes requested):

docker/release/build_all.sh
yohaimagen: I don't think it is overkill. For my use case, using Tandem on a Kubernetes cluster is preferable, so the container can be as lean as possible. A single compiled Tandem is ~400 MB, the entire compiled Tandem compiler is ~800 MB.  In the documentation, I focus solely on the all-in-one container.
It will be created only on a new version release, so it is only once in a very long while. 
 I'm happy to remove it if you still think it is unnecessary (it is easy for me to make them specifically for my use case). Let me know.

docs/getting-started/docker-release.rst
yohaimagen: In the docker run command, I bound `/host` to the host `pwd` as suggested in the link. 

I added the following warning at the end of the tutorial to inform users to save simulation output on the host: 


Tandem writes output files to the path specified in the .toml configuration. If /host is not bound to your host machine as shown in the mount step above, all output will be written inside the container and lost when it exits. Simulation output files can be very large; writing them inside the container may also exhaust the container’s writable layer and cause unexpected failures.

docker/release/Dockerfile.base
yohaimagen: It should not be used directly. Installing a Python env with pytest, pandas, vtk, and mshio is unnecessary for the release version Docker image, for example.  
@piyushkarki, if we would like, you can create a new `Dockerfile.base` that includes only the bare minimum base requirements for a Tandem installation. 
PETSc
Eigen
Tandem source code 
similarly to https://github.com/TEAR-ERC/tandem/blob/ymagen/docker-release-images/docker/release/Dockerfile.base and place it under /docker/base 
Then we can both use it for our own purposes.
And I will change this PR to branch out of  
https://github.com/TEAR-ERC/tandem/tree/piyush/thomas/docker

.github/workflows/release-images.yml
yohaimagen: I'm leaving this one open. I will remove it after everything else is resolved.

docs/getting-started/docker-release.rst
yohaimagen: I added gmsh to the release images. and edited the docs accordingly.

.github/workflows/release-images.yml
yohaimagen: Agreed, changed to true.

docker/release/Dockerfile.all
yohaimagen: I dont think it is very essential for a release image as those will be created once in a long while. It is important for the test images. Nevertheless, I added ninja compilation.

docker/release/Dockerfile.base
Thomas-Ulrich: ok, I think we want to keep the Python env Python env  on the test base docker.
And I agree at least the python vtk can take a few 100Mb, so if the aim is to have a lean dockerfile, then let's have 2 dockerbase.

docker/release/build_all.sh
Thomas-Ulrich: "A single compiled Tandem is ~400 MB, the entire compiled Tandem compiler is ~800 MB."
Alternatively, you could keep only tandem and static in each folder, which would reduce a lot the size of the docker? I think it would be better to have only one container, but if you want to keep all containers, I'm also fine with that.
I also wonder if GitHub will allow us to run such a long job building the full container. Probably there are time limitations for a free account.

Thomas-Ulrich:
also @piyushkarki, any idea why the CI fails?

piyushkarki:
@Thomas-Ulrich looking at the CI error, it seems there are some PETSc syntax errors so I suspect using 3.24.5 is not yet possible because of the PETSc syntax changes that have not been adopted into tandem yet.

docker/release/Dockerfile.base
piyushkarki: Okay I will have two docker files. One in base that only has the bare minimum dependencies for tandem installation. The test docker will build on this base docker file and will add a virtual environment and it's own python setuptools/dependencies for the pytests. Does that make sense to you? Also will try to adopt ninja. Tried it locally and it "feels" faster for sure.

yohaimagen: @piyushkarki sounds good to me

docker/release/build_all.sh
yohaimagen: ok, after keeping only the `tandem` and `static` executables, the image sizes are pretty comparable. I removed all 12 single tandem images. 
changed the nameing convention in the relase image to `tandem_xd_yp` 'static_xd_yp` and updated the docs. 
The build time is ~30 minutes. GitHub has not caused any trouble with building it so far.

docker/release/Dockerfile.base
piyushkarki: Okay it's done. Now you will find two docker files in PR #97 under docker/base and docker/test. The base image has only the bare minimum required for tandem installation. The test image takes this base image and builds on top of it the python requirements and later can be used to further change things if required (with good reason).

The workflows set up in #97 only allows the image building and pushing to registry if:

1. The base or test docker file has changed
2. A new PETSc 3.22 patch is released
3. Someone has deleted  the image from the registry

Otherwise, it will just pull from the correctly tagged version from the registry and build tandem in it.

yohaimagen: Thanks @piyushkarki for setting up the split base/test Dockerfiles.

**What was done on this PR:**

- Dropped `docker/release/Dockerfile.base` entirely, the release image now builds directly on top of `ghcr.io/tear-erc/tandem/tandem-base`
- Removed the `build-base` CI job from this PR's workflow
- Rebased this branch onto `piyush/thomas/docker`
- The only addition needed on top of the shared base was `python3-numpy`, required by the yateto kernel generation script at build time (not included in the base image) but could be included I think @piyushkarki  and then removed here

**@piyushkarki** — the `:latest` tag for `tandem-base` is not yet available on GHCR. The workflow change to add it is in place, but it needs to be triggered (e.g. via a manual workflow dispatch or a push that touches the base Dockerfile). Once `:latest` is live, this PR will switch from the hardcoded `3.22.5` tag to `:latest` as the final step.

**@Thomas-Ulrich** — there are old test image tags left over from earlier iterations of this PR that should be cleaned up from the `tandem` package on GHCR. Could you either delete them or grant me package admin rights at https://github.com/orgs/TEAR-ERC/packages/container/tandem/settings?

Tags to delete:
- `all-ymagen-docker-release-images`
- `2d-1p-ymagen-docker-release-images` through `3d-6p-ymagen-docker-release-images` (12 tags)

The only tag to keep is `ymagen-docker-release-images` (the current all-in-one build).

Thomas-Ulrich: I don't have permissions to administrate the package or to remove docker images.

yohaimagen: @AliceGabriel, I need permission to finalize this PR
Could you either delete them and make the final image version public or grant me package admin rights


Delete unnessery images:
There are leftover test image tags from earlier iterations of this PR that should be removed from the tandem package on GHCR. 
https://github.com/orgs/TEAR-ERC/packages/container/tandem/settings

Tags to delete:
all-ymagen-docker-release-images
2d-1p-ymagen-docker-release-images through 3d-6p-ymagen-docker-release-images (12 tags)
The only tag to keep is ymagen-docker-release-images (the current all-in-one build). 


Change to public:
Go to https://github.com/orgs/TEAR-ERC/packages/container/tandem/settings
scroll to Danger Zone → Change visibility → Public
This is a one-time action — all future tags pushed to these packages will automatically inherit public visibility.

.github/workflows/release-images.yml
yohaimagen: Done, I cleaned the workflow to be triggered only on a version release.

yohaimagen:
@piyushkarki @Thomas-Ulrich 

This is ready to be merged downstream to Piyus's PR. 

Note on ARM CPUs. I managed to create a working image built on my laptop, but was not able to create one that can be built on GitHub and run on my Mac M3. For now, I've put all those attempts on a new bench (ymagen/arm-release-images) and will revisit them later when I have more time.

Thomas-Ulrich:
Ok, I won't have time to look at it again because of the force-push, which prevents me from identifying the new changes.

Thomas-Ulrich (approved):
LGTM

PR #98, opened 2026-03-15
Volume tagging

piyushkarki:
## Summary:
This PR introduces volume tagging support, allowing material properties to be assigned regionally based on integer tags defined in the mesh (physical groups in Gmsh). The element tags are parsed, preserved during MPI parallel mesh partitioning, and exposed directly to Lua configuration scripts, enabling region-aware material property functions for Elasticity and Poisson operators.

## Description: Changes
- Mesh & Geometry: Captured element tags during mesh parsing in GlobalSimplexMeshBuilder (addElement). Introduced VolumeData (inheriting from MeshData) to handle the storage and MPI redistribution (via AllToAllV) of these tags across partitions. Extended Curvilinear to store and provide access to these tags via getVolumeTag(elNo).

- Lua Scripting Integration: Extended LuaLib with getMemberFunctionTagged and functional_t_region to allow Lua material functions to accept an additional tag argument alongside spatial coordinates.

- Form & Operators: Updated DGCurvilinearCommon::make_volume_functional to evaluate material functions by passing both the quadrature coordinates and the element's volume tag. ElasticityScenario, PoissonScenario, and SeasScenario were updated to use functional_t_region for material properties (mu, lam, rho, K). These can all be unified under a single functional_t once facet tagging comes in.

- CI/Tests: Updated GitHub Actions to fetch test data from the volume-tagging-test branch of tandem_test_data. Refactored existing Python parallel consistency tests to be reusable and added a new volume tagging consistency test.

## Tests
- Unit tests:
  - VolumeData round-trip: Verified that volume tags are correctly packed, redistributed across MPI ranks using AllToAllV, and unpacked identically.
  - Out-of-range lids: Verified that redistribution safely assigns a default tag of -1 for out-of-range local IDs.
  - make_volume_functional: Created a  2-triangle mesh with tags 1 and 2, evaluated a mock material function that switches on the tag, and verified the resulting quadrature matrix exactly matches the expected branch values.

- Integration tests:
  - Parallel Consistency: Added test_parallel_consistency_volume_tagging.py to ensure that VTU outputs produced by multi-rank partitioned runs with volume tags are strictly consistent with single-rank runs.
- Correctness test: Added [test_correctness_volume_tagging.py](https://github.com/TEAR-ERC/tandem/blob/refs/heads/piyush/volume-tagging/test/test_correctness_volume_tagging.py) to check for correctness of volume tagging using a 2D Poisson problem.

## Additional Info
If a mesh is loaded that does not provide volume tags (or an older builder is used), the system safely defaults to assigning a tag of -1 to all elements and prints a warning to std::cerr.

## tldr:
Volume tags from the mesh are now passed to Lua scripts. You can define region-specific materials in your TOML/Lua config like this: 
```
function mu(x, y, tag) 
  if tag == 1 then 
    return 30.0 
  else 
    return 40.0 
  end 
end
```


test/reference_configs/2D/volume_tagging_parallel_consistency.geo
```diff
@@ -0,0 +1,43 @@
...
+
+// This call below to BooleanFragments will result in 16 line segments being defined
+// The original lines are deleted.
+// It will also result in 4 volumes being defined - and the original volume will be deleted.
+v() = BooleanFragments{ Surface{box}; Delete; }{ Line{h_line, v_line}; Delete; };
+
+
+MeshSize{ PointsOf{Surface{:};} } = res;
+
+//Physical Curve(1) = { 1, 8, 9, 11, 12, 6, 7, 4 }; // exterior
```
Thomas-Ulrich: remove commented lines?

test/reference_configs/2D/volume_tagging_parallel_consistency.lua
```diff
@@ -0,0 +1,50 @@
...
+function VolumeTagging:new(o)
+    o = o or {}
+    setmetatable(o, self)
+    self.__index = self
+    return o
+end
+
+
+function VolumeTagging:mu(x, y, tag)
+    local _tag = math.floor(tag)
```
Thomas-Ulrich: why do you need to use math.floor?
In any case, if you want to avoid floor(3.99999999) =  3
you may want to try:
`local _tag = math.floor(tag + 0.5)`

test/reference_configs/2D/volume_tagging_parallel_consistency.lua
```diff
@@ -0,0 +1,50 @@
...
+function VolumeTagging:lam(x, y)
+    return 2.0
+end
+
+function VolumeTagging:force(x, y)
+    return 5.0*math.pi^2*math.cos(math.pi*x)*math.cos(math.pi*y),
+          -3.0*math.pi^2*math.sin(math.pi*x)*math.sin(math.pi*y)
+end
+
+function VolumeTagging:solution(x, y)
```
Thomas-Ulrich, line 35: do you really need solution and solution_jacobian in this file?
(or just copy paste from another example). Then I would remove if not needed

src/mesh/MeshData.h
```diff
@@ -139,6 +139,39 @@ class ElementData : public MeshData {
     NumberingConvention convention_;
 };
 
+class VolumeData : public MeshData {
```
Thomas-Ulrich: calling it VolumeTagData would avoid confusion, as VolumeData are generally stored as ElementalData

Thomas-Ulrich (approved):
LGTM (as far as I can understand). See minor comments nevertheless.

baroryan:
I am able to compile and run, thanks for implementing this feature! 

Maybe one thing to consider is to add an example to show users how to use the feature:
for example change mu function in examples/2d/bp3.lua with (and call the new lua bp3_voltag.lua)
```
function BP3:mu(x, y,tag)
        local rho
        if tag == 1 then
                rho=self.rho0/2
        else
                rho=self.rho0
        end
    return self.cs^2 * rho
end
```

and then another bp3_voltag.toml that refers to the new bp3_voltag.lua

baroryan:
also one quick question when I use this:

```
function BP3:mu(x, y,tag)
        local rho
        if tag == 1 then
                print(tag)
                rho=self.rho0/2
        else
                rho=self.rho0
                print(tag)
        end
    return self.cs^2 * rho
end
```





I am getting a lot nil results from the lua (though I should only get 1/2 as defined in the mesh )

for example: 

2
2
2
nil
nil
nil
nil


--- Is Tandem calling this function for fault elements where tag is not defined ? Could this cause some ill defined scenarios on the lua end ?

src/mesh/MeshData.h
piyushkarki: I like this. Was also iffy about the name. Thanks! Reminds me of [this](https://martinfowler.com/bliki/TwoHardThings.html#:~:text=There%20are%20only%20two%20hard%20things%20in%20Computer%20Science%3A%20cache%20invalidation%20and%20naming%20things).

test/reference_configs/2D/volume_tagging_parallel_consistency.geo
piyushkarki: Removed.

test/reference_configs/2D/volume_tagging_parallel_consistency.lua
piyushkarki: This is something I inherited from an old PR. Here: https://github.com/TEAR-ERC/tandem/pull/39#:~:text=I%20note%20that%20you%20have%20defined%20the%20region%20index%20as%20a%20float%2E

We will not have this issue since it's always long int. I will remove the floor. Thanks!

test/reference_configs/2D/volume_tagging_parallel_consistency.lua
piyushkarki: I would need the solution for the boundary but yes I removed the solution_jacobian.

piyushkarki:
@baroryan  thank you for taking a look! 

1. I will add a BP3 copy with modified function signatures for mu and lambda.

2. For the ````nil``` cases: My guess is, you get this because you are probably calling mu from other lua functions as well, such as lambda, where you might be calling mu without a tag in which case this is expected (please let me know if that is not the case). So, you could add tag to lambda as well but then you need to propagate this tag into your mu call too. Now I am thinking if there are functions which require mu but are never fed with a tag. Would be nice to know. It's only mu, rho and lambda for this PR since "volume" tags only apply to these bulk quantities. For the fault tagging PR, you can see almost all functions are fed with a tag so this won't be an issue.

baroryan:
Thanks Piyush! You were right about 2, my bad : )

docs/first-model/parameters.rst
```diff
@@ -7,6 +7,7 @@ Tandem simulation parameters are written in the toml script. Here are the key pa
 - **mesh_file**: Mesh file.
 - **lib**: Lua file containing material & frictional paramters.
 - **scenario**: Name of the specific scenario defined in the Lua library.
+- **Lua material functions**: When volume tagging is used, `mu`, `lam`, and `rho` Lua functions also receive the volume tag as the last argument.
```
Thomas-Ulrich: What about:
Note that the mu, lam, and rho Lua functions receive the volume tag as the last function argument. If region IDs are not set in the mesh, a value of -1 is assumed.

test/reference_configs/2D/volume_tagging_parallel_consistency.lua
```diff
@@ -10,7 +10,7 @@ end
 
 
 function VolumeTagging:mu(x, y, tag)
-    local _tag = math.floor(tag)
+    local _tag = tag
```
Thomas-Ulrich: in the end you don't need _tag its seems, you could use directly tag?

Thomas-Ulrich (approved):
LGTM now. See also new comments.

test/reference_configs/2D/volume_tagging_parallel_consistency.lua
piyushkarki: Yes wanted to keep it because I liked it stylistically but removing this for consistency. Also in an example lua file, added an example where you can use tag directly.

docs/first-model/parameters.rst
piyushkarki: Adapted this. Thanks.

PR #99, opened 2026-03-17
HDF5 probe writers and moment rate writer

piyushkarki:
## Summary:
This PR introduces parallel HDF5 output writing for tandem, specifically adding hierarchical output for moment rates, fault probes, and domain probes. It implements the necessary mathematical formulations for moment rate evaluation, HDF5 parallel I/O writers, and configuration extensions to allow saving simulation time-series data in .h5 format.

## Description: Changes
- Kernel & Operator Formulations: Added moment_rate and slip_rate evaluation to both Elasticity and Poisson adapters. Added mu_avg (average shear modulus) calculation for quadrature point interpolation.
- Friction Operator Updates: Integrated moment_rate_local() calculation into the friction operator, which interpolates slip rate values at basis nodes to quadrature nodes using the adapter.
- Parallel HDF5 Writers: Implemented core parallel HDF5 I/O using HDF5Writer, HDF5ProbeWriter, and MomentRateWriter. Uses the collective MPI writes (H5FD_MPIO_COLLECTIVE), hyperslabs, and extendible datasets (H5Dset_extent) for time-series data.
- App & Config Integration: Added TOML schema blocks for HDF5_moment_rate_output, HDF5_fault_probe_output, and HDF5_domain_probe_output. Handled DataLevel::Hierarchical in the SEAS monitor to trigger writes at active time steps.
- Fallbacks: If ENABLE_HDF5=OFF, dummy implementations throw runtime errors, and the SEAS initializer prints a warning if HDF5 outputs were requested in the configuration.
Docker/CI Integration: Added h5py to the Python testing dependencies.

## Tests
- HDF5Writer Unit Tests: 
    - Check initialization, dataset creation, and dataset extension across multiple timesteps
    - Verify files are properly closed and resource handles are released on destruction. 
    - Verify correct error throwing when HDF5 support is disabled.

- Integration tests:
    - Moment Rate Consistency: Verify ```moment_rate.h5``` is correctly formatted (fault_elements x timesteps x 1) and that the normalized sum of moment rates at the first timestep strictly matches the expected value.
    - Probe Output Consistency: Python pytest (```test_probe_writer_consistency.py```) verifies that the data dumped to HDF5 fault probes (```fltst.h5```) and domain probes (```blkst.h5```) exactly matches the data written by the legacy CSV probe writers across all time steps and fields.

## tldr:
Compile with ```-DENABLE_HDF5=ON``` and define the requested outputs in the ```TOML``` script using:

 ```[moment_rate_output]``` => Available only in HDF5 format no need to specify type
 
 
 ```
 [fault_probe_output]
 type = "csv" <------------------Only this is new
 ...
 (same prefix, location, frequency as previous writers)
 ```
 
 
 or,
 
 
 ```
 [domain_probe_output]
 type = "csv"<------------------Only this is new
 ...
 (same prefix, location, frequency as previous writers)
 
 ``` 
 
 with probe locations just like in CSV outputs to write parallel time-series output to HDF5.


hpc4geo:
@piyushkarki Similar to my comment to PR#97, I would like HDF5 to not become a hard dependency. Could you please add support in CMAKE to allow HDF5 to be optional (maybe with a flag such as TANDEM_HAVE_HDF5 for example). 

Happy to discuss other options, or arguments why every tandem must compile and link against HDF5 from now on.

.github/workflows/build-dependencies.yml
```diff
@@ -90,6 +90,7 @@ jobs:
         run: |
           if docker manifest inspect "${{ inputs.dependencies_image }}:${{ inputs.petsc_version }}-hdf5${{ inputs.hdf5_version }}" > /dev/null 2>&1; then
             echo "exists=true" >> $GITHUB_OUTPUT
+            echo "image_exists=true" >> $GITHUB_OUTPUT
```
Thomas-Ulrich: what is the difference between 
exists and image_exists?

app/localoperator/Elasticity.cpp
```diff
@@ -1016,4 +1016,14 @@ void Elasticity::traction_boundary(std::size_t fctNo, FacetInfo const& info,
...
+void Elasticity::mu_avg(std::size_t fctNo, FacetInfo const& info, Matrix<double>& result) const {
+    assert(result.size() == tensor::mu_avg_q::size());
+
+    kernel::mu_avg krnl;
+    krnl.mu_q(0) = fctPre[fctNo].get<mu_q_0>().data();
+    krnl.mu_q(1) = fctPre[fctNo].get<mu_q_1>().data();
+    krnl.mu_avg_q = result.data();
+    krnl.execute();
+}
+
```
Thomas-Ulrich, line 1033: Does it really make sense to use yateto for adding 2 vectors?

```
void Elasticity::mu_avg(std::size_t fctNo, FacetInfo const& info, Matrix<double>& result) const {
    double const* mu0 = fctPre[fctNo].get<mu_q_0>().data();
    double const* mu1 = fctPre[fctNo].get<mu_q_1>().data();
    double* res = result.data();

    #pragma omp simd
    for (std::size_t q = 0; q < nq; ++q) {
        res[q] = 0.5 * (mu0[q] + mu1[q]);
    }
}
```

test/reference_configs/2D/bp1_ref_QDGreen_probe_comparison_hdf5.toml
```diff
@@ -0,0 +1,49 @@
...
+]
+
+[HDF5_fault_probe_output]
+prefix = "../../temp_test_results/fltst"
+t_max = 9460800
+probes = [
+    { name = "dp000", x = [0.0, 0.0] },
+]
+
+[domain_probe_output]
```
Thomas-Ulrich, line 24: Wouldn't it be simpler to have 
[domain_probe_output]
format= 'csv' or 'hdf5'?

app/kernels/elasticity_adapter.py
```diff
@@ -25,3 +26,14 @@ def add(generator, dim, nbf_fault, nq):
...
+    slip_rate = Tensor('slip_rate', (nbf_fault, dim))
+    slip_rate_q = Tensor('slip_rate_q', (dim, nq))
+    moment_rate = Tensor('moment_rate', (dim,))
+
+    generator.add(
+        'evaluate_slip_rate', slip_rate_q['pq'] <=
+            e_q['lq'] * slip_rate['lp']) 
+    generator.add(
+        'evaluate_moment_rate', moment_rate['p'] <=
+            w['q'] * slip_rate_q['pq'] * nl_q['q'] * mu['q'])
```
Thomas-Ulrich: Can't this be evaluated in one step?
```
generator.add(
    'evaluate_moment_rate', moment_rate['p'] <= 
        w['q'] * e_q['lq'] * slip_rate['lp'] * nl_q['q'] * mu['q']
)
```

app/form/FrictionOperator.h
```diff
@@ -103,7 +109,28 @@ template <typename LocalOperator> class FrictionOperator : public AbstractFricti
             }
 
             VMax_ = std::max(VMax_, VMax);
+
+            // Interpolate slip rate values at Basis function nodes to quadrature
+            // nodes
+            if (adapter_) {
+                std::vector<double> slip_rate_q_raw(LocalOperator::NumQuantities * quadRuleSize_);
```
Thomas-Ulrich: shouldn't you allocate this memory once and for all ourside the
`for (std::size_t faultNo = 0, num = num_local_elements(); faultNo < num; ++faultNo) {`
loop?

app/form/FrictionOperator.h
```diff
@@ -103,7 +109,28 @@ template <typename LocalOperator> class FrictionOperator : public AbstractFricti
             }
 
             VMax_ = std::max(VMax_, VMax);
+
+            // Interpolate slip rate values at Basis function nodes to quadrature
+            // nodes
```
Thomas-Ulrich: probably this comment should be 
// Compute moment rate for each quadrature point

app/form/FrictionOperator.h
```diff
@@ -218,10 +245,13 @@ template <typename LocalOperator> class FrictionOperator : public AbstractFricti
 
 private:
     std::unique_ptr<LocalOperator> lop_;
+    std::unique_ptr<AbstractAdapterOperator> adapter_;
     std::shared_ptr<DGOperatorTopo> topo_;
     std::shared_ptr<BoundaryMap> fault_map_;
     Scratch<double> scratch_;
     double VMax_ = 0.0;
+    std::vector<double> moment_rate_ = {};
```
Thomas-Ulrich: Wouldn't a  `std::vector<std::array<double, D-1>>` be more readable and intuitive?

app/form/AdapterOperator.h
```diff
@@ -77,6 +77,19 @@ template <typename LocalOperator> class AdapterOperator : public AbstractAdapter
         result.end_access(result_handle);
     }
 
+    void slip_rate(Vector<double const>& slip_rate_reshaped, Matrix<double>& slip_rate_q) override {
+        lop_->slip_rate(slip_rate_reshaped, slip_rate_q);
+    }
+    void moment_rate(std::size_t faultNo, Matrix<double>& moment_rate_vector,
```
Thomas-Ulrich: I see that in `void traction(BlockView const& displacement, BlockVector& result) override {` above, the loop 
`for (std::size_t faultNo = 0, num = num_local_elements(); faultNo < num; ++faultNo) {`
is in the function. Is it not possible to do the same here?

app/tandem/Writer.h
```diff
@@ -229,6 +242,131 @@ template <std::size_t D> class DomainWriter : public Writer {
...
+    MomentRateWriter(std::string_view prefix, AdaptiveOutputInterval oi,
+                     LocalSimplexMesh<D> const& mesh, std::shared_ptr<Curvilinear<D>> cl,
+                     unsigned degree, BoundaryMap const& bnd_map, MPI_Comm comm)
+        : Writer(prefix, oi), writer_(prefix, comm),
+          adapter_(mesh, std::move(cl), bnd_map.localFctNos(), degree), degree_(degree),
+          comm_(std::move(comm)) {}
+
+    DataLevel level() const override { return DataLevel::Hierarchical; }
+    bool has_static_writer() const override { return true; }
+    void write(double time, std::vector<double> data) override {
```
Thomas-Ulrich: probably you want to pass data by reference:
const std::vector<double>& data

app/tandem/Writer.h
```diff
@@ -229,6 +242,131 @@ template <std::size_t D> class DomainWriter : public Writer {
...
+template <std::size_t D> class MomentRateWriter : public Writer {
+public:
+    MomentRateWriter(std::string_view prefix, AdaptiveOutputInterval oi,
+                     LocalSimplexMesh<D> const& mesh, std::shared_ptr<Curvilinear<D>> cl,
+                     unsigned degree, BoundaryMap const& bnd_map, MPI_Comm comm)
+        : Writer(prefix, oi), writer_(prefix, comm),
+          adapter_(mesh, std::move(cl), bnd_map.localFctNos(), degree), degree_(degree),
+          comm_(std::move(comm)) {}
+
+    DataLevel level() const override { return DataLevel::Hierarchical; }
```
Thomas-Ulrich: actually I don't understand why it is not boundary

test/test_probe_writer_consistency.py
```diff
@@ -0,0 +1,155 @@
...
+    try:
+        with h5py.File(h5_file_path, "r") as h5:
+            # Check expected datasets exist and are non-empty
+            for key in ["time", "probeData", "probeFields", "probeNames"]:
+                assert key in h5, f"[{prefix}] HDF5 missing dataset '{key}'"
+                assert h5[key].size > 0, f"[{prefix}] HDF5 dataset '{key}' is empty"
+
+            h5_time = h5["time"][:]
+            h5_data = h5["probeData"][:]
+            h5_fields = [
```
Thomas-Ulrich: according to AI, this should also work:
`h5_fields = h5["probeFields"].asstr()[:]`

test/test_probe_writer_consistency.py
```diff
@@ -0,0 +1,155 @@
...
+    except OSError as e:
+        raise AssertionError(
+            f"moment_rate.h5 could not be opened — possibly corrupted or not "
+            f"properly closed (e.g. simulation crashed mid-write): {e}"
+        )
+
+    # Validate shape: (num_fault_elements, num_timesteps, 1)
+    assert moment_rate.ndim == 3, (
+        f"momentRate has unexpected shape {moment_rate.shape}, "
+        f"expected (fault_elements, timesteps, 1)"
```
Thomas-Ulrich, line 34: - Seems like a lot of data stored! Could it be a mode in which you just store the integrated moment rate release over all the fault (one time serie), or over each segment (if fault tag enabled)?
- Also, it seems the moment rate could be a field of the fault output, actually (that could be visualized in Paraview) (or let's say a new VTU fault output if you want it outputted at a different frequency).
- Also I note that it uses the same adaptive time sampling output as the other outputs. So basically could be printed at much lower frequencies (which would make less sense?). In that sense, it would make more sense to print seismic moment, instead of seismic moment rate (that the user could later on differentiate in post-processing)?
- Does the new output works with checkpointing?

test/HDF5Writer.cpp
```diff
@@ -0,0 +1,105 @@
...
+            std::vector<hsize_t> initial_dims = {1, 3}; // timestep x 3
+            std::vector<hsize_t> max_dims = {H5S_UNLIMITED, 3};
+            int extensibleDim = 0;
+            int glueDim = 0;
+
+            hid_t dset = writer.createExtendibleDataset(
+                "time_series", H5T_NATIVE_DOUBLE, initial_dims, max_dims, extensibleDim, false);
+            CHECK(dset >= 0);
+
+            std::vector<double> timestep0 = {0.1, 0.2, 0.3};
```
Thomas-Ulrich, line 32: I think it is a bit confusing:
are you writing data like a probe data field or is it the time?
Then change timestep0 by a better name e.g. data0


And in:
```
            writer.writeToDataset(dset, H5T_NATIVE_DOUBLE, 0, timestep0.data(), write_dims, glueDim,
                                  extensibleDim, false);
```

the 0 is the time value (double) ?
In this case maybe change to 0 and 1e-3 to make it clear.

Maybe it's also best to test with 2 elements, not 1, so it more closely resembles the true dataset.
(I mean, I think unit tests are also good for showing other developers how your code works)

test/CMakeLists.txt
```diff
@@ -58,9 +58,20 @@ add_executable(test-util util.cpp)
...
+    find_package(HDF5 COMPONENTS C HL QUIET)
+endif()
+
+if(ENABLE_HDF5 AND HDF5_FOUND)
+    add_executable(test-H5Parser H5Parser.cpp)
+    target_link_libraries(test-H5Parser test-runner-mpi)
+    doctest_discover_tests(test-H5Parser)
+endif()
+
+add_executable(test-HDF5Writer HDF5Writer.cpp)
```
Thomas-Ulrich, line 67: why not under 
if(ENABLE_HDF5 AND HDF5_FOUND)
?

test/scripts/2D/HDF5_probe_writer.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
Thomas-Ulrich, line 1: maybe rename script to:
generate_hdf5_probe_test_output.sh

src/io/HDF5Writer.h
```diff
@@ -0,0 +1,48 @@
...
+#define HDF5_WRITER_H
+
+#include <cstdint>
+#include <mpi.h>
+#include <string>
+#include <string_view>
+#include <tuple>
+#include <vector>
+
+#ifdef ENABLE_HDF5
```
Thomas-Ulrich, line 11: Why not having all the class under #ifdef ENABLE_HDF5?

test/CMakeLists.txt
```diff
@@ -58,9 +58,20 @@ add_executable(test-util util.cpp)
 target_link_libraries(test-util test-runner)
 doctest_discover_tests(test-util)
 
-add_executable(test-H5Parser H5Parser.cpp)
-target_link_libraries(test-H5Parser test-runner-mpi)
-doctest_discover_tests(test-H5Parser)
+if(ENABLE_HDF5)
+    find_package(HDF5 COMPONENTS C HL QUIET)
```
Thomas-Ulrich: See prefered parallel in SeisSol
https://github.com/SeisSol/SeisSol/blob/master/CMakeLists.txt#L206-L224

Thomas-Ulrich (changes requested):
See comments below.
(You can consider those requiring the most refactoring as suggestions).

.github/workflows/build-dependencies.yml
piyushkarki: Just a remnant of a rebase. Removing this. Thanks.

app/form/AdapterOperator.h
piyushkarki: Couldn't see the bigger picture in the past. Adopted this pattern now. Thanks!

app/form/FrictionOperator.h
piyushkarki: Changed - not relevant here anymore.

app/form/FrictionOperator.h
piyushkarki: Changed the pattern. This is now similar.

app/form/FrictionOperator.h
piyushkarki: The write function takes a flat structure so I kept it as such. Faster than repackaging I suppose.

app/kernels/elasticity_adapter.py
piyushkarki: Yes slip_rate had been useful initially for debugging but removing this altogether and fusing these kernels now.

app/localoperator/Elasticity.cpp
piyushkarki: Yes also another step where I missed the bigger picture. Converted it to direct calculation now. Thank you!

app/tandem/Writer.h
piyushkarki: I just wanted one at the time to distinguish from other DataLevels so I could separate the data type while developing. Now changed it to DataLevel::Area just to distinguish it other data types for moment rate but since now I fuse the HDF5_probe_writers with probe_writers, the DataLevel are going to be boundary for those.

app/tandem/Writer.h
piyushkarki: Yup changed to const ref data type. Thanks.

src/io/HDF5Writer.h
piyushkarki: The class needs to stay visible at compile time because we want this to compile even when ENALBE_HDF5 is off. So by design choice, we don't want the class itself to depend on the HDF5 library but only the implementation inside and then we throw the proper errors when someone tries to use the instantiation. That is, always declare the class, stub the methods with a std::runtime_error in the #else branch. Also what was suggested here: https://github.com/TEAR-ERC/tandem/pull/97#issuecomment-4077192782

test/reference_configs/2D/bp1_ref_QDGreen_probe_comparison_hdf5.toml
piyushkarki: Wasn't simpler from where I see it but was the right design I feel. I agree. Changed the implementation to follow this.

test/scripts/2D/HDF5_probe_writer.sh
piyushkarki: Just sticking to the unit test files naming convention where the files are just named according the to files being tested.

test/CMakeLists.txt
piyushkarki: I want these to compile unconditionally so even without HDF5 enabled, I would like to see the code throw appropriate errors.

test/HDF5Writer.cpp
piyushkarki: Yes it is for time 0. I will change this to 1e-3 to make things clearer. Thanks!

I do add two elements, timestep0 and 1. These are essentially two chunks of data. Could you clarify what do you mean by elements?

test/test_probe_writer_consistency.py
piyushkarki: 1. This feature was actually motivated by the requirements of the BP7 benchmark, where moment rate output is needed for spatially varying regions of the fault. Previously, the only option was to generate multiple VTU outputs and then post-process them (e.g., filter regions and integrate slip rates with vtk/pyvista/paraview tools). This approach was both:

    - computationally expensive (due to full VTU writes), and
    - less accurate (compared to integration using quadrature weights and basis functions directly).

The current implementation provides a more efficient and accurate alternative by storing a relatively “thin” dataset that can later be filtered spatially (and, in the future, by fault tags).

2. I agree that including moment rate as part of the fault VTU output (potentially with a separate output frequency) would be cool, especially for visualization.

However, since this involves changes to the VTU output pipeline and possibly different output scheduling, I would suggest handling this as a separate feature/issue.

3. Since we are calculating both slip and slip rate, computationally, it would be similar but since the quantity of interest was the moment rate, we wrote this specifically. I assume moment is more robust to time changes since it does not require a differentiation in time. Is that why you suggest storing moment as compared to moment rate? Please tell me if I misunderstood. I would add though that we should make it another issue that we can work on after this release. 

4. I didn't consider checkpointing before but now I have integrated it into the writers and added tests for them. Thank you!

test/test_probe_writer_consistency.py
piyushkarki: Changed this to get array properly with list comprehension.

JeenaYun:
@piyushkarki, my small test suggests that HDF5 and CSV probe outputs are comparable. Both formats yield the same output dimension (i.e., the number of time and location points) as it should be, and the differences between variables from HDF5 and CSV probe outputs at each probe location are <= O(1e-15) for on-fault variables and ~O(1e-10) for domain variables (`u0` & `u1`). My test was done in 2D, polynomial degree 6, aging law, `type=elasticity`, and `mode=QDGreen`. I attached the input files I used to test and the figures that show the differences: [hdf5_probe_test.tar.gz](https://github.com/user-attachments/files/26560687/hdf5_probe_test.tar.gz)

My only suggestion is to print statements such as "HDF5 probe output enabled" during the configuration stage when `-DENABLE_HDF5=ON`, just like the friction law selection (see L54:64 in `CMakeLists.txt`). This would avoid confusion.

test/test_probe_writer_consistency.py
Thomas-Ulrich: ok!
about 3., if the moment rate calculation is triggered only at each moment-rate output writing, (defined by atol, rtol,etc, and which could be much larger than dt of the simulation), then moment rate can be less representative of the fully time-sampled moment rate (because depending on local SR), while seismic moment depends on slip (which is integrated at each dt).
So let's say if you want to compute seismic moment from seismic moment rate, it may differ from the true seismic moment.
(in the end, it depends on what quantity we want to get precisely:  seismic moment or moment rate).

about 2, I agree of handling this as a separate feature/issue.

src/io/HDF5Writer.cpp
```diff
@@ -6,15 +6,31 @@ namespace tndm {
...
+
     // Create HDF5 file with parallel access
     hid_t plist_id = H5Pcreate(H5P_FILE_ACCESS);
     H5Pset_fapl_mpio(plist_id, comm_, MPI_INFO_NULL);
-    file_ =
-        H5Fcreate((std::string(filename) + ".h5").c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, plist_id);
+
+    // Check if file exists and checkpointing is enabled
+    if (enable_checkpoint) {
+        htri_t file_exists = H5Fis_hdf5(filepath.c_str());
```
Thomas-Ulrich, line 22: I wonder what would happen if the last run was interrupted during the last output write...
Probably the file would be corrupted. 
(maybe open an issue and fix this issue in a future PR).

src/io/HDF5Writer.cpp
```diff
@@ -6,15 +6,31 @@ namespace tndm {
 
 #ifdef ENABLE_HDF5
 
-HDF5Writer::HDF5Writer(std::string_view filename, MPI_Comm comm) : comm_(comm) {
+HDF5Writer::HDF5Writer(std::string_view filename, MPI_Comm comm, bool enable_checkpoint)
```
Thomas-Ulrich: checkpoint_enabled?

app/tandem/Writer.h
```diff
@@ -36,7 +36,7 @@ enum class DataLevel {
     Volume
 #ifdef ENABLE_HDF5
     ,
-    Hierarchical
+    Area // Fault surface data (moment rate) computed over fault area
```
Thomas-Ulrich: Fault surface data (moment rate) integrated over each fault element (one scalar per fault element) ?
Or is it one scalar per quadrature point?
Maybe: BoundaryForMomentRate

test/reference_configs/2D/bp1_ref_QDGreen_probe_comparison_hdf5.toml
```diff
@@ -14,36 +14,21 @@ storage_type="none"
...
 [fault_probe_output]
-prefix = "../../temp_test_results/fltst_"
-t_max = 9460800
-probes = [
-    { name = "dp000", x = [0.0, 0.0] },
-]
-
-[HDF5_fault_probe_output]
 prefix = "../../temp_test_results/fltst"
+type = "HDF5"
```
Thomas-Ulrich, line 18: maybe you could even have a type="both" enabling both writers, allowing to run only one simulation.
But anyway, that would only save you one simulation in the CI.

test/HDF5Writer.cpp
```diff
@@ -98,8 +98,122 @@ TEST_CASE("HDF5Writer - dataset extends correctly") {
...
+            writer.createExtendibleDataset("data", H5T_NATIVE_DOUBLE, dims, max_dims, 0, false);
+
+        for (int t = 0; t < 3; ++t) {
+            std::vector<double> data = {t * 10.0, t * 20.0};
+            writer.writeToDataset(dset, H5T_NATIVE_DOUBLE, t, data.data(), {1, 2}, 0, 0, false);
+        }
+        writer.closeDataset(dset);
+    }
+
+    // Reopen WITHOUT checkpoint - should OVERWRITE (fresh file)
```
Thomas-Ulrich, line 210: actually, best would be to copy the behaviour of seissol, that is rename all previous outputs, not to loose them, e.g.:
`dyn_0025_coh0.25_1.0_B0.9_C0.5_R0.6-fault_vertex.h5` to 
`dyn_0025_coh0.25_1.0_B0.9_C0.5_R0.6-fault_vertex.bak_2025-11-16_11-13-49.h5`

(but I would understand that this is outside the range of the PR. In that case, maybe you want to open an issue feature request to implement that latter? Or that someone else implement that latter.)

Thomas-Ulrich (approved):
Great that you followed my suggestions, even if they implied a lot of refactoring, thank you!
Overall, the PR should alreaddy merge ready, by see added comments:

(and there is also this comment, that you ignored: See prefered parallel in SeisSol
https://github.com/SeisSol/SeisSol/blob/master/CMakeLists.txt#L206-L224)

JeenaYun:
@piyushkarki My last comment would be documenting the usage. [PR101](https://github.com/TEAR-ERC/tandem/pull/101) updates documentation regarding installation. Do you think you can merge those updates and add 1) installation for parallel-enabled HDF5 libraries, 2) compilation of this branch (both in `docs/getting-started/installation.rst`), and 3) a guide on how to use the HDF5 writer feature as you did in `tldr` in `docs/first-model/parameters.rst`.

src/io/HDF5Writer.cpp
piyushkarki: Done

src/io/HDF5Writer.cpp
piyushkarki: Yes will start an issue for this

test/HDF5Writer.cpp
piyushkarki: Yes will start an issue on this.

app/tandem/Writer.h
piyushkarki: It is a surface data yes. It is the area integral over a fault element for each direction so it is D-1 scalars per fault element. But I agree Area and boundary are ambiguous. Replacing with BoundaryForMomentRate. Thanks!

test/reference_configs/2D/bp1_ref_QDGreen_probe_comparison_hdf5.toml
piyushkarki: I think having both would not make sense here. The previous implementation actually could do both together because you had different schema name altogether. But with this round of review, I think it makes more sense to just have one. You wouldn't want two different outputs at once unless it's for testing, which you can do by having two sections with different type in each and compiling with HDF5.

test/CMakeLists.txt
piyushkarki: Will try to adopt this after the merge so I can do this together for all HDF5 dependencies.

